PR #41, opened 2014-07-14
WIP: Add Filters

dereuromark:
It might be wise to add some common and generic filters in /csv/Filter as a collection of often used filters.
This is more DRY than having to add them for each project.
This also increases the usage and quality of filters.


nyamsprod:
in the next major version the following filters will be bundle

- CharsetConverter : to convert using the mbstring extension CSV document
- RFC4180Field: to improve CSV document interoperability

PR #67, opened 2015-01-31
Use parent methods in MapIterator

codeliner:
The \IteratorIterator accepts any \Traversable object not only Iterators. To support \Traversable objects like PDOStatements as inner iterator you need to access the current item via the iterator methods of the outer iterator.

### DIFF
--- a/src/Iterator/MapIterator.php
+++ b/src/Iterator/MapIterator.php
@@ -52,6 +52,6 @@ class MapIterator extends IteratorIterator
         $iterator = $this->getInnerIterator();
         $callable = $this->callable;
 
-        return $callable($iterator->current(), $iterator->key(), $iterator);
+        return $callable(parent::current(), parent::key(), $iterator);
     }
 }


nyamsprod:
Hi @codeliner , thanks for the PR.
 I'm just wondering since the MapIterator is an internal class which is never exposed in the public API. And the only class which uses it is the Reader class which only interacts with Iterators objects what would be the package gain of making the MapIterator working with Traversable object ?

codeliner:
Hi @nyamsprod, I use a similar MapIterator in [my library](https://github.com/gingerwork/gingerframework/blob/master/library/Ginger/Functional/Iterator/MapIterator.php) (inspired by your version) and ran into trouble with it. I thought it is a good idea to fix the same problem in your version even if it is only used internally.
We are working with PHP so no package visibility. Sure you have this internal flag in the doc block but this doesn't prevent my IDE to suggest me your MapIterator when the package is installed. 
Long story short, it is no real improvment for your lib, just a hint. It does not affect the behaviour, but if you don't want to merge it then it is fine, too.

Btw. thanks for this great library. I use it in my [workflow system](https://github.com/gingerwork/app-core/blob/master/module/FileConnector/src/Service/FileTypeAdapter/LeagueCsvTypeAdapter.php). I love the iterator filters.

nyamsprod:
@codeliner Just for your information in the upcoming v7 the MapIterator has moved places. Another reason to not use class tagged internal :+1: . That being said I first merge your PR but I had to rollback to the previous implementation due to scrutinizer not liking your version of MapIterator

codeliner:
@nyamsprod Thanks. No problem. Like mentioned before I use my own version of the map iterator. I'm wondering that scrutinizer didn't like my changes. What was the problem?
However, it's working for you. Just keep in mind that your implementation won't work with all kinds of \Traversable objects but it excepts all \Traversable objects as valid constructor arguments. For me the internal usage would not be a valid reason for a error-prone implementation but if it is more important for you that scrutinizer like it then it's fine, too.

Issue #78, opened 2015-03-12
Typo in Docs


zanderbaldwin:
Typo of "tree" instead of "three" in Reading section of docs. PR coming soon. Not particularly exciting but every little helps!

PR #79, opened 2015-03-12
Fix Typo in Reading Section of Docs

zanderbaldwin:
Fix typo "tree" to "three". Fixes issue #78.


nyamsprod:
thanks, these are the kind of typo I can not catch :+1:

Issue #83, opened 2015-03-28
Filtering out the BOM


bogdanghervan:
[The docs](http://csv.thephpleague.com/bom/#getinputbom) say that we should rely on the extracting methods to remove the BOM character while the CSV is read, but that doesn't seem to be the case.

Even though the BOM could be detected, it is not automatically removed, but rather included in the first column read.

I've put together an example here to illustrate what I'm doing:
http://runnable.com/VRYJNJ-c0wNVqTc6/parsing-bom-character-for-php

Am I missing something, or is it really up to the developer to manually remove it?

nyamsprod:
Hi,
thanks for using the package. I'm afraid that yes it is up to the developer to manually remove it. As the  `Reader` can not modify the CSV. you can do that using for instance the callable argument of any `Reader::fetch*` methods.

nyamsprod:
I understand your concern but stripping even a single bite of a CSV document with the `Reader` class is simply not acceptable. As per definition a `Reader` class **should not be allowed** to change the nature of the object it interacts with. So in our case we're left with two options:

- **option 1:** As stated in the documentation. Use the extract methods `$callable` argument to remove them.
- **option 2:**: Re-download your CSV and remove the BOM sequences using the BOM methods.

I understand completely that dealing with BOM sequences can be a pain, but unless we come out with a solution that respect the `Reader` contract. I'm afraid the current behavior will stay.

Richtermeister:
I am not implying that the reader should alter the underlying file it reads. Clearly that would be unacceptable. 

However, just like the reader produces a transformed representation of that file (turning delimited to array, filtering, sorting, limiting, etc) I think it would be a simple matter of registering one more internal filter, to strip the BOM character from the first field of the first row before handing this data back to userland code. As it is the Reader registers several Iterator decorators already, this would just be one more.

nyamsprod:
I see your point but your are mistaking the underlying mechanism. the decorators helps selecting rows, They do not in anyway interacts with them. The only way to remove effectively the BOM sequence is through the use of the `$callable`. But in the current codebase this is not possible maybe you can submit a PR and I'll review it. Like I said I'm open for any solution that do not alter the CSV document.


- This filtering can only occurs throught the use of the callback argument 
Remember that all filters are cleared after each query which means that you'll have to register it every time you filter your document.

nyamsprod:
The next major release (v9.0) will automatically remove the BOM on CSV document read. The patch landed on the master branch.

PR #86, opened 2015-04-03
[WIP] Strip BOM Character

Richtermeister:
Proposed solution for https://github.com/thephpleague/csv/issues/83

Let me know what you think.

### DIFF
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -72,6 +72,13 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
     protected $open_mode;
 
     /**
+     * Whether to automatically strip the BOM character
+     *
+     * @var bool
+     */
+    protected $strip_bom = false;
+
+    /**
      * Csv Controls Trait
      */
     use Config\Controls;
@@ -142,6 +149,15 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
         }
         $iterator->setCsvControl($this->delimiter, $this->enclosure, $this->escape);
         $iterator->setFlags($this->flags);
+
+        if ($this->strip_bom) {
+            $this->strip_bom = false; // avoid recursive call to getIterator()
+            $bom = $this->getInputBOM();
+            $this->strip_bom = true;
+            if (! is_null($bom)) {
+                $iterator = new Modifier\StripBomIterator($iterator, $bom);
+            }
+        }
 
         return $iterator;
     }
@@ -259,6 +275,7 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
         $csv->input_bom    = $this->input_bom;
         $csv->output_bom   = $this->output_bom;
         $csv->newline      = $this->newline;
+        $csv->strip_bom    = $this->strip_bom;
 
         return $csv;
     }
@@ -285,5 +302,19 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
     public function newReader($open_mode = 'r+')
     {
         return $this->newInstance('\League\Csv\Reader', $open_mode);
+    }
+
+    /**
+     * Set whether to automatically strip BOM character
+     *
+     * @param bool $flag
+     *
+     * @return $this
+     */
+    public function stripBom($flag)
+    {
+        $this->strip_bom = (bool) $flag;
+
+        return $this;
     }
 }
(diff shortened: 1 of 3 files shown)


nyamsprod:
This is a good start. I think you could improve the PR by:

- adding a property and its setter/getter methods that would default to not removing the BOM sequence so that we do not introduces meaningless BC break

src/Reader.php
```diff
@@ -52,6 +52,20 @@ public function query(callable $callable = null)
         $iterator = $this->applyIteratorFilter($iterator);
         $iterator = $this->applyIteratorSortBy($iterator);
         $iterator = $this->applyIteratorInterval($iterator);
+
```
nyamsprod: How about moving the code prior to `applyIteratorFilter` line ? This is an open question I don't know if it would change something or not ?

nyamsprod: By moving it Up you ensure that the BOM sequence will be indeed remove from the First line of the document. After the filters have been applied you may be remove a valid BOM sequence inside the CSV

Richtermeister: Ah yes, we want to respond to the original set, not the filtered one. Will implement.

Richtermeister:
Style question, are you ok with this code inline, or would you prefer a separate StripBomModifier, to enhance readability/testability?

nyamsprod:
I would prefer the code being added on a StripBomModifier but that customisation :) let's make the thing work first :+1: 

- test
- code (PSR-2)

then at the end re-organization if needed

nyamsprod:
@Richtermeister are you still working on this feature or not ?

Richtermeister:
Yes, apologies, got distracted. Will push update tonight.

Richtermeister:
WIP - Still working on tests.

src/AbstractCsv.php
```diff
@@ -259,6 +275,7 @@ protected function newInstance($class_name, $open_mode)
         $csv->input_bom    = $this->input_bom;
         $csv->output_bom   = $this->output_bom;
         $csv->newline      = $this->newline;
+        $csv->strip_bom    = $this->strip_bom;
```
nyamsprod: I believe the `strip_bom` property should not be attached to `AbstractCsv` class. This property should be only usable on the `Reader` class. 

IMHO to make the code as much flexible as possible you should add the strip bom capability directly in the `Modifier/QueryFilter.php` trait and reset the status to false between calls. 

**Pros:**
- the BOM stripping is restricted to the `Reader` class
- the BOM stripping can be set per queries (so stripping is only made when needed).

**Cons:**
- the BOM stripping can be set per queries. The dev must not forget to use it

Secondly keep in mind that the `fetchAssoc` method can be affected if the `strip_bom` is set to `true`  and the column names are taken from the first row (which is the default behavior).

nyamsprod:
@Richtermeister If you are too busy I can take over this feature it won't be too difficult for me as you already laid out the foundation for a solution, thought ?

Richtermeister:
@nyamsprod Absolutely, have at it, thank you.

Just to explain the approach I took.. I was trying to implement the BOM-stripping very transparently across all csv access functions (once explicitly enabled via stripBom(true)), including the fetchOne method.. this is why I tried to customize the behavior of the getIterator method. The reason is that getRow() is used when you work with associative keys, and this is where my original problem with the BOM character occurred.

Let me know if this makes sense or not :) Happy to elaborate. And thank you for the help, I really enjoy this library, you did a great job on it.

nyamsprod:
I've just push my modification to the master branch I'll close this pull request. You can pull from the master branch to see if it works like you want.

nyamsprod:
When you have time could you review the changes I've made to the package to add your feature please.
I've already updated the [documentation](http://csv.thephpleague.com/reading/#modifying-content-methods)

- If you have other wishes this is the time to make them or I'll considered 7.1 feature/fix completed
- you can also look at the [CHANGELOG](https://github.com/thephpleague/csv/blob/master/CHANGELOG.md) to see the others changes in the package.

Richtermeister:
@nyamsprod Yes, awesome, I will check it out on the weekend.

nyamsprod:
@Richtermeister did you have some time to review the feature... I'm waiting for your review to release 7.1

Richtermeister:
@nyamsprod Yes, I did. We have a winner! :+1: Nice work man, works exactly right. Thank you very much!

nyamsprod:
version 7.1 is released with the stripBom method

Richtermeister:
Very cool! Upgraded :+1:

PR #87, opened 2015-04-03
Add associative iterator

Richtermeister:
This is just an idea I had while using this lib on large imports. I like the ability to have associative keys, but the build-in functionality forces this to read the entire file failed due to memory limits.

This proposed solution allows us to enable associative keys, yet lazily apply them to each row during iteration.

### DIFF
--- a/src/Modifier/AssociativeIterator.php
+++ b/src/Modifier/AssociativeIterator.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * This file is part of the League.csv library
+ *
+ * @license http://opensource.org/licenses/MIT
+ * @link https://github.com/thephpleague/csv/
+ * @version 7.0.1
+ * @package League.csv
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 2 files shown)


nyamsprod:
Could you show me a example code on how this works ?

nyamsprod:
I think the next iteration of the package will throw away `arrays` in favor of `generator`. I could not do it before because PHP5.5 adoption rate was not good but since PHP 5.4 will be EOL is 6 months or so when the new version will be release this problem will be resolved itself.

Richtermeister:
@nyamsprod I apologize for not being very responsive, as I am very busy at work, but I would like to pursue this idea further and I believe it makes your library better and more user-friendly. It's ok if this functionality will be obsoleted in 6 months, but that is still a long time down the road. Would you mind keeping this ticket open for me?

nyamsprod:
@Richtermeister no problem but first just send me a example so I can understand the implication as a gist for example. As I said this can be dealt with Generator in a next major release or using the Reader::query method using a well written callable.

PR #109, opened 2015-06-25
Empty enclosures should be allowed as per RFC4180

dakira:
RFC4180 lists enclosures as a SHOULD requirement which allows for them to be omitted. Some legacy apps are not able to handle CSV with enclosed strings so users should be able to not use enclosures.

### DIFF
--- a/src/Config/Controls.php
+++ b/src/Config/Controls.php
@@ -159,8 +159,8 @@ trait Controls
      */
     public function setEnclosure($enclosure)
     {
-        if (1 != mb_strlen($enclosure)) {
-            throw new InvalidArgumentException('The enclosure must be a single character');
+        if (1 < mb_strlen($enclosure)) {
+            throw new InvalidArgumentException('The enclosure must be a single character or an empty string');
         }
         $this->enclosure = $enclosure;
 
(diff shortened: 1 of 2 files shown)


nyamsprod:
@dakira thanks for submitting your pull request. Regarding the enclosure delimiter I think the package already follow the RFC4180 which states

> Due to lack of a single specification, there are considerable differences among implementations.  Implementors should "be conservative in what you do, be liberal in what you accept from  others"

The Writer class just does that by enforcing the use of an enclosure to ensure that the output CSV always have a enclosure defined for interoperability.

Furthermore, PHP's function and method to create a CSV document requires an enclosure to be set. If no enclosure is set, the default enclosure `"` will be use. So your patch won't work at all.

To work around this PHP "limitation" (which is not IMHO) I've seen people use the `\0` byte character as the enclosure character. This hack which already works on the package assumes:

- that no null byte are used inside the CSV field. Because if a null byte is present you will corrupt the CSV data by introducing double null byte (see RFC4180).
- the software that will read the generated CSV trim/not take into account any null byte sequence in the CSV data. Again this can lead to data corruption.

**TL;DR:** Even thought, the enclosure character is optional most of the time in CSV export, it is required in some situations. Adding the ability to create an enclosure-less CSV would degrade the package quality and interoperability.

dakira:
I understand. Thanks for taking so much time to reply! I guess in my edge-case I'll just use a distinct enclosure and filter that out while I ask the software vendor to patch their stuff.

PR #121, opened 2015-10-12
Don't trim $filter_name in sanitizeStreamFilter

HeyRatFans:
When using `appendStreamFilter` to add a filter to a Writer (or Reader) instance, the name of the filter is sanitised by the `sanitizeStreamFilter` method in StreamFilter.php. This method casts the name to a string, trims and returns it.

Trimming the string causes issues when implementing a filter where it's last argument has a trailing new-line character, for example I am attempting to write a filter that uses `str_replace` which will be used to convert Windows-style line endings (\r\n) to Linux-style (\n), but the Filter's onCreate method receives only the first argument due to trim removing the tailing new line.

**Example code:**

`$writer = Writer::createFromPath( tempnam( sys_get_temp_dir(), 'csv' ) );
stream_filter_register( FilterReplace::FILTER_NAME . "*", FilterReplace::class );
$writer->appendStreamFilter( FilterReplace::FILTER_NAME . "\r\n:\n" );`

Thanks.

### DIFF
--- a/examples/lib/FilterReplace.php
+++ b/examples/lib/FilterReplace.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace lib;
+
+use php_user_filter;
+
+class FilterReplace extends php_user_filter
+{
+    const FILTER_NAME = 'convert.replace.';
+
+    private $search;
+
+    private $replace;
+
+    public function onCreate()
+    {
+        if( strpos( $this->filtername, self::FILTER_NAME ) !== 0 ){
+            return false;
+        }
+
+        $params = substr( $this->filtername, strlen( self::FILTER_NAME ) );
+
+        if( !preg_match( '/([^:]+):([^$]+)$/', $params, $matches ) ){
+            return false;
+        }
+
+        $this->search  = $matches[1];
+        $this->replace = $matches[2];
+
+        return true;
+    }
+
+    public function filter( $in, $out, &$consumed, $closing )
+    {
+        while( $res = stream_bucket_make_writeable( $in ) ){
+
+            $res->data = str_replace( $this->search, $this->replace, $res->data );
+
+            $consumed += $res->datalen;
+
+            /** @noinspection PhpParamsInspection */
+            stream_bucket_append( $out, $res );
+        }
+
+        return PSFS_PASS_ON;
+    }
+}
--- a/src/Modifier/StreamFilter.php
+++ b/src/Modifier/StreamFilter.php
@@ -203,9 +203,7 @@ trait StreamFilter
     protected function sanitizeStreamFilter($filter_name)
     {
         $this->assertStreamable();
-        $filter_name = (string) $filter_name;
-
-        return trim($filter_name);
+        return (string) $filter_name;
     }
 
     /**
--- a/test/StreamFilterTest.php
+++ b/test/StreamFilterTest.php
@@ -4,6 +4,7 @@ namespace League\Csv\test;
 
 use League\Csv\Reader;
 use League\Csv\Writer;
+use lib\FilterReplace;
 use PHPUnit_Framework_TestCase;
 use SplFileObject;
 use SplTempFileObject;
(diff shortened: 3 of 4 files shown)


nyamsprod:
hi @HeyRatFans thanks for submitting your PR but I think that you are dealing with 2 problems:

* converting your CSV newlines (this problem is for the Csv::writer class)
* creating a good stream filter name

- For the first problem you could simply use the [setNewline](http://csv.thephpleague.com/properties/#the-newline-sequence) method :) . And as far as I am aware PHP's SplFileObject has no problem handling correctly the newline sequence while reading the CSV files, unless you are using a Mac based computer or a CSV that was generated on a Mac based computer for which you can simply set a [specific ini value](csv.thephpleague.com/instantiation/#mac-os-server).

- For the second problem PHP stream filter name have some restrictions for instance using "new line characters or "invisible" ones in them is forbidden" so using `\r` or `\n` or even `\0` (the null byte) will trigger a Fatal error when using them. If you had included some tests with your PR like it is recommend by the [contribution guidelines](https://github.com/thephpleague/csv/blob/master/CONTRIBUTING.md) you would have stumble on this issue for sure.

TL;DR: If you still want to use stream filter to achieve your goal you need to think of an workaround to specify you input but for those reasons I can not accept your PR.

HeyRatFans:
Hi,

Isn't the setNewLine method used to specify the line-terminator? This isn't intended to change that, but instead alter the text in the fields, i.e.: `1,"two",3,"new\r\nline"` would become `1,"two"3,"new\nline"` (note the change in new line the 4th field), which we cannot otherwise change using Writer.

We need to do this to normalise the new-lines in the data we export, which unfortunately we can't alter at source and for speed I am using PDO and passing a statement directly into Writer, which is what attracted me to this library in the first place.

Do you have a source for using invisible characters in a filter name? I had no problem creating a stream filter which does exactly as described above with the modified StreamFilter.php I have submitted and I did not encounter any errors, fatal or otherwise.

Apologies for not contributing a test, I have now added one to test the filter name, but not the actual functionality. Do you need one for this?

Thanks.

nyamsprod:
Okay so you want to change the line terminators **inside** the cell okay got it :) 
here's a blog post I made about enforcing CSV delimiters using stream filters. Hope you'll find what you want in there.
http://nyamsprod.com/blog/2015/qa-enforcing-enclosure-with-leaguecsv/
IMHO in your case you should create a "simple" filter with a fixed filtername. Since you already know what you want to change and with what you want to change it with. The code will be much simpler

HeyRatFans:
I had seen your blog entry previously, which actually gave me the idea of using a StreamFilter in the first place! :)

I did originally start with a simpler filter that did just line-endings, but I have other use cases where a generic str_replace stream filter would be more useful than multiple classes and not just with this library.

nyamsprod:
I understand what you are trying to do but the main issue is not really League\Csv. To illustrate, I'll use your stream filter and `SplFileObject`:

```php
use lib\FilterReplace;

require 'FilterReplace.php';

stream_filter_register(FilterReplace::FILTER_NAME.'*', FilterReplace::class);
$streamPath = 'php://filter/read='.FilterReplace::FILTER_NAME."\r\n:\n"."/resource=".__DIR__."/test.csv";
new SplFileObject($streamPath); //SplFileObject will throw a RuntimeException
```

* I've register your filter
* `League\Csv` will keep a registry of the registered filter (this is what your added test just verify)
* League\Csv uses lazy loading, so the SplFileObject is only created when needed (when using `foreach`) . On `foreach` call  the object generates the `streamPath` and feed it to SplFileObject constructor since this is [the only way to apply stream filtering to a SplFileObject](https://bugs.php.net/bug.php?id=44392). 
* Because you've added `\r`, `\n` and other characters that have a numerical value < 32 `SplFileObject` will throw a `RuntimeException` because you've submitted an invalid path.

**TL;DR:** Removing the path trimming will **not** resolve you issue.

HeyRatFans:
What exception are you're seeing? I _was_ seeing this:

`PHP Fatal error:  Uncaught exception 'RuntimeException' with message 'SplFileObject::__construct(): unable to create or locate filter "convert.replace.\r\n:\n"'`

Which was entirely my mistake as I hadn't commit the latest version of the FilterReplace class. The version submitted previously wasn't correctly matching "\r\n:\n" which caused the onCreate method to return false and the above Exception to be raised.

I'm not getting any other exceptions with your code snippet now.

Thanks.

HeyRatFans:
I have updated the tests to show that the stream filter does work to replace \r\n with \n with `Writer`. No exception is throw by writing the file! I have double checked this on both Ubuntu and Windows 7.

However, I have discovered an interesting issue using `Reader` to read back the written data. If I use PHP's native fopen and fgetcsv I can verify the data is written successfully, but using `Reader` the data in my "new line" field appears to be stripped of **all** new lines, including both \r\n and \n and I can't determine where or why this is occurring. This happens regardless of whether FilterReplace is attached to the Reader.

Hopefully you'll know better :)

Thanks.

nyamsprod:
Could you please:
- Open an issue related to this PR
- Make a PR from one of your a dedicated branch and not from your master branch.
- Remove the fgetcsv test which does not test the package per se.
- Once this is done I'll accept your PR but I reserve the right to change/adapt some of your code to make it pass the HHVM.
- Open a distinct issue for the one you just found so I can work on it separately.

Except for the tests changes, what I'm asking is written in the CONTRIBUTING to make it easier for any maintainer and future users to quickly find references to bugs in the issue tracker. 

As your patch boils down to removing the usage of the `trim` function on the stream filter name I'm ok with it :+1: .

HeyRatFans:
Closing in favour of #123.

PR #130, opened 2015-10-21
Bug Fix Spl Flags

nyamsprod:
- default flag is now SplFileObject::READ_AHEAD|SplFileObject::SKIP_EMPTY
- AbstractCsv::__toString is also bug fixed


nyamsprod:
This PR fixes two bugs related to SplFileObject flags used in the library
- The default flags are now SplFileObject::READ_AHEAD|SplFileObject::SKIP_EMPTY. So that using foreach will return the expected data even when multiple lines are presents in CSV fields
- the __toString method needs to normalize the flag usage to output correclty the file. Flags were badly applied which led to missing rows in the output

Issue #72, opened 2015-02-20
library doesn't support iconv stream filter


nsitbon:
Hi, I just pulled the latest version of the library and try to integrate it in my application but it fails at the beginning when I do
```php
$stream = Reader::createFromPath('/home/nsitbon/export_members_20150217.csv');

if ($stream->isActiveStreamFilter()) { 
    $stream->appendStreamFilter('convert.iconv.UTF-16/UTF-8//TRANSLIT');
}
```
it throws the following exception : 
> [RuntimeException]
>   SplFileObject::__construct(): unable to create or locate filter "convert.iconv.UTF-16"

whereas this code works perfectly
```php
$stream = fopen('/home/nsitbon/export_members_20150217.csv', 'r');
$stream_filter_append($stream, 'convert.iconv.UTF-16/UTF-8//TRANSLIT', STREAM_FILTER_READ);
```

any ideas?

nyamsprod:
This is a limitation due to poor support from `SplFileObject` of stream filters. There's a [open bug](https://bugs.php.net/bug.php?id=44392) about this and it has not yet been resolve.

nyamsprod:
In the meantime, nothing prevents you from modifying the [FilterTranscode](https://github.com/thephpleague/csv/blob/master/examples/lib/FilterTranscode.php) class to use iconv extension instead. I understand it is frustrating but if this bug was resolved a even more powerful package could be made and It would also remove the need for `isActiveStreamFilter` in the package.

nsitbon:
thanks for your comments, if you can't do anything about that I'll wait for the fix in the SPL.
I spend some time in the documentation but cannot find if this behaviour is documented, can you tell me if there is something about this in the documentation. I'm sure there is nothing in the tests because I checked.

nyamsprod:
Yep there's nothing, a omission from my part. But you can make a pull request against the [gh-pages](https://github.com/thephpleague/csv/blob/gh-pages/filtering.md) branch to add the warning.

nyamsprod:
I've just updated the documentation by adding a warning section on top of the stream filter documentation page.

nsitbon:
thank you much Ignace.

zerocrates:
FYI, the `php://filter` stream filter support *does* appear to work with filters that contain slashes, as long as you percent-encode the slashes, i.e.

`php://filter/read=convert.iconv.UTF-8%2FASCII%2F%2FTRANSLIT/resource=...`

[...]

PR #152, opened 2016-02-09
Bugfix Library

nyamsprod:
- bug fix stream filter #72
- improve public API by deprecating (get|set)EncodingFrom

### DIFF
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -190,7 +190,7 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
         $csv->delimiter = $this->delimiter;
         $csv->enclosure = $this->enclosure;
         $csv->escape = $this->escape;
-        $csv->encodingFrom = $this->encodingFrom;
+        $csv->input_encoding = $this->input_encoding;
         $csv->input_bom = $this->input_bom;
         $csv->output_bom = $this->output_bom;
         $csv->newline = $this->newline;
--- a/src/Config/Output.php
+++ b/src/Config/Output.php
@@ -33,7 +33,7 @@ trait Output
      *
      * @var string
      */
-    protected $encodingFrom = 'UTF-8';
+    protected $input_encoding = 'UTF-8';
 
     /**
      * The Input file BOM character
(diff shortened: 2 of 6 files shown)


src/AbstractCsv.php
```diff
@@ -190,7 +190,7 @@ protected function newInstance($class, $open_mode)
         $csv->delimiter = $this->delimiter;
         $csv->enclosure = $this->enclosure;
         $csv->escape = $this->escape;
-        $csv->encodingFrom = $this->encodingFrom;
+        $csv->input_encoding = $this->input_encoding;
```
hannesvdvreken: just curious: why snake_case instead of camelCase?

nyamsprod: I was in the mood for snake case :) . Seriously I did this to match $input_bom and $output_bom which are already in snake case. I should probably review all the codebase and make a clear decision between both coding style. Of note none of them are in contradiction to PSR2 if I recall but yes the code needs consistency.

hannesvdvreken: > I was in the mood for snake case
:smile: 

> none of them are in contradiction to PSR2
You're right

src/Config/Output.php
```diff
@@ -54,26 +54,56 @@
...
 
         return $this;
     }
 
     /**
+     * Sets the CSV encoding charset
+     *
+     * DEPRECATION WARNING! This method will be removed in the next major point release
+     *
+     * @deprecated deprecated since version 4.1
```
vlakoff: since version 4.1? should be 8.x.x (to be released)

src/Config/Output.php
```diff
@@ -54,26 +54,56 @@
...
+    {
+        return $this->input_encoding;
+    }
+
+    /**
      * Gets the source CSV encoding charset
      *
+     * DEPRECATION WARNING! This method will be removed in the next major point release
+     *
+     * @deprecated deprecated since version 4.1
```
vlakoff: same here

src/Config/Output.php
hannesvdvreken: It's deprecated sinde 4.1, and will be removed in the next major version.

vlakoff: I just thought of some typo / old PR. Okay then, I trust you :)

nyamsprod:
@vlakoff indeed you are correct it is a typo and the deprecation message was changed in the master. :+1: 

I should point out that I had a hard time finding this discussion since the comment was made on an already closed PR

Issue #157, opened 2016-03-23
BOM length is appended to the end of the csv output


Lidbetter:
Looking through RFC's I can't find where this is listed.

When calling `$writer->output('filename');`

the output file looks like this:
csv:
![screen shot 2016-03-23 at 12 39 36](https://cloud.githubusercontent.com/assets/6525616/13993047/f56dd586-f0f4-11e5-9b03-5d3f677b98b5.png)
numbers:
![screen shot 2016-03-23 at 12 38 58](https://cloud.githubusercontent.com/assets/6525616/13993050/fa098fcc-f0f4-11e5-862b-2e54e875426a.png)

nyamsprod:
Hi @Lidbetter this question was already raised and answered to on the following issue #107 .

PS: what RFC are you referring to ?

Lidbetter:
Thanks, I was returning the output from my controller, rather than just calling output

Issue #107, opened 2015-06-12
output() appends a row with the filesize?


JoeDawson:
Hey!

Excellent package by the way.

Currently outputting my CSV to discover that there's a random row being added to the end which appears to be the filesize.

```php
use League\Csv\Writer as CSV;

$csv = CSV::createFromFileObject(new \SplTempFileObject());

$csv->insertOne(['Venue Name', 'Total Entries', 'Passed', 'Failed']);

foreach($report['venues'] as $venue)
{
	$total =  $report['passes'][$i]+ $report['fails'][$i];

	$csv->insertOne([$venue->name, $total, $report['passes'][$i], $report['fails'][$i]]);

	$i++;
}

return $csv->output('report-'. $request->user()->id .'-'. Carbon::now()->timestamp .'.csv');
```

This outputs the following table;

![Output](http://f.cl.ly/items/2Q1g0i0m3d1x2R3s041n/Screen%20Shot%202015-06-12%20at%2013.02.53.jpg)

What's odd, is that when I use `$csv->toHTML()`, I just get this?

![toHTML](http://f.cl.ly/items/0322281B3H1P273M1Y0w/Screen%20Shot%202015-06-12%20at%2013.11.31.jpg)

Any ideas why the filesize is being appended when using `output()`?

Thanks in advance!

nyamsprod:
Hi @JoeDawson,
Thanks for using the library. Like stated in the [documentation for the output method](http://csv.thephpleague.com/basic-usage/)  
> Since version 7.0, the method returns the number of characters read from the handle and passed through to the output.

This was done to match [PHP native function fpassthru](http://php.net/fpassthru)

The real question is why are you using `output` to save your CSV, this method is meant to be used to download the CSV, no `return` is needed 99% of the time :) .

JoeDawson:
Oops, thanks @nyamsprod!

Wasn't aware that the `return` wasn't needed. Thank you very much!

PR #158, opened 2016-03-23
Add more detailed description to the behaviour of output

Lidbetter:
Might help prevent more tickets like these:
#157 
#107

### DIFF
--- a/src/Config/Output.php
+++ b/src/Config/Output.php
@@ -168,7 +168,11 @@ trait Output
     abstract public function getIterator();
 
     /**
-     * Outputs all data on the CSV file
+     * Outputs all data on the CSV file. 
+     * The CSV data  will be written to the output buffer. Returning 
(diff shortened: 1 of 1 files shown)


nyamsprod:
The return type is clearly documented using docblock.

What I would rather propose is improve the documentation around this method on the documentation website by making a PR against the gh-pages.

src/Config/Output.php
```diff
@@ -168,7 +168,11 @@ public function getInputBOM()
     abstract public function getIterator();
 
     /**
-     * Outputs all data on the CSV file
+     * Outputs all data on the CSV file. 
+     * The CSV data  will be written to the output buffer. Returning 
+     * the result of this function (or anything else) may append
```
nyamsprod: This description adds more confusion ... there's not **may** !! The method signature [is very strict](http://csv.thephpleague.com/basic-usage/#output)

PR #166, opened 2016-06-03
Improve performance by moving the LimitIterator to the top

piwi91:
Move the limit iterator to the top because it isn't using `seek` because other iterators aren't implementing the `SeekableInterface`. This is a problem when using an offset.

# Profiling with Blackfire

Profile 1: current implementation
Profile 1

Profile 2: my implementation (moved the `LimitItterator`)
Profile 2

### DIFF
--- a/src/Modifier/QueryFilter.php
+++ b/src/Modifier/QueryFilter.php
@@ -158,10 +158,10 @@ trait QueryFilter
         };
         array_unshift($this->iterator_filters, $normalizedCsv);
         $iterator = $this->getIterator();
+        $iterator = $this->applyIteratorInterval($iterator);
         $iterator = $this->applyBomStripping($iterator);
         $iterator = $this->applyIteratorFilter($iterator);
         $iterator = $this->applyIteratorSortBy($iterator);
-        $iterator = $this->applyIteratorInterval($iterator);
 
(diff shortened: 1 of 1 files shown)


nyamsprod:
@piwi91 You can not simply move the limitIterator to the top as you will loose the "SQL like" nature of the query. 

In SQL you first make the selection and then you apply the limits. your PR is doing the opposite, you first apply a limit and then make selection on it.

nyamsprod:
Looking at Spl documentation I would say that the 2 culprits are ArrayObject and CallbackFilterIterator.

- `ArrayObject` can be replace by `ArrayIterator` and inherit the `SeekableInterface`. This change is trivial.

-  `CallbackFilterIterator` is the real buttlenek as we can not simply remove or replace it and by definition this class can not implement the SeekableInterface unless you can come up with a better alternative.

nyamsprod:
@piwi91 I've changed ArrayObject -> ArrayIterator into the master branch. Did you find another solution for CallbackFilterIterator or could you close this PR ?

piwi91:
@nyamsprod I have used another solution to import the CSV data (I load the data into the MySQL database first, this is done in a few seconds).

I will close this PR

PR #171, opened 2016-08-01
Change getInputBom() method name to getInputBOM() to be consistent everywhere in the API.

hhamon:


nyamsprod:
thanks for the PR 👍

PR #175, opened 2016-09-01
Change Output header Content-Type to text/csv

mahngiel:
## Introduction

Currently, `Output#output` expects to render a CSV file, but does not set the correct Content-Type header.

Per [RFC 7111](https://tools.ietf.org/html/rfc7111), the content type is available `text/csv`. 

Further, [HTTP 1.1 7.2.1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7) states `octet-stream` should only be used when the content-type is uknown. 
> If and only if the media type is not given by a Content-Type field, the recipient MAY attempt to guess the media type via inspection of its content and/or the name extension(s) of the URI used to identify the resource. If the media type remains unknown, the recipient SHOULD treat it as type "application/octet-stream". 

`Output#output` sets Content-Type as `text/csv`

None

### Targeted release version

8.1.1

### PR Impact

None observed

## Open issues

None

### DIFF
--- a/src/Config/Output.php
+++ b/src/Config/Output.php
@@ -179,7 +179,7 @@ trait Output
     {
         if (null !== $filename) {
             $filename = filter_var($filename, FILTER_SANITIZE_STRING, FILTER_FLAG_STRIP_LOW);
-            header('Content-Type: application/octet-stream');
+            header('Content-Type: text/csv');
             header('Content-Transfer-Encoding: binary');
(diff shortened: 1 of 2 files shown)


nyamsprod:
@mahngiel 
Could you submit the CONTRIB changes in another PR please as these changes should be decoupled with the current PR ?

mahngiel:
@nyamsprod done. see https://github.com/thephpleague/csv/pull/176

Issue #202, opened 2017-01-23
Please use resources internally instead of SplFileObject instances.


frankdejonge:
## Issue summary

Due to the need to use SplFileObject instances internally any input it required to be compatible with the restrictions said class puts in them. Every SplFileObject method we call has a function counterpart. The only thing we actually use is the iterative capabilities of the object. This could/should be replaces with an own implementation.

If we accomplish this we achieve a higher level of interoperability, we're able not only to work with local files but also with resources from remote locations, which is, by any definition, simple "the shit".

----------

### System informations

| Information | Description |
|--------------|---------|
| League\Csv version | next |
| PHP/HHVM version | any |
| OS Platform | dna |

## Standalone code, or other way to reproduce the problem

Try to use the existing package with a resource and weep.

### Expected result

More interoperability

### Actual result

A sad face because it's not yet at this level of awesomeness.

nyamsprod:
This issue is resolved in the master branch ... rejoice my friend as the answer is being review for release

PR #203, opened 2017-01-24
Adding createFromStream #202

nyamsprod:
This commit adds the ability to work with resource stream by adding a StreamIterator object that mimick SplFileObject and takes a stream resource as its unique constructor argument. 
To work as intended the resource stream **needs to be seekable**.

```php
<?php

$csv = Reader::createFromStream(fopen('php://temp', 'r'));
```

This now will work for the `Reader` and the `Writer` and @frankdejonge will no longer make a sad face


src/AbstractCsv.php
```diff
@@ -127,6 +128,35 @@ public static function createFromFileObject(SplFileObject $file)
...
+     *
+     * @return static
+     */
+    public static function createFromStream($stream)
+    {
+        if (is_resource($stream)) {
+            $stream = new StreamIterator($stream);
+        }
+
+        if ($stream instanceof StreamIterator) {
```
frankdejonge: I'd suggest to negate this check and throw this block, keeping the happy path alive.

nyamsprod: yes indeed ... /me over checking 👍

nyamsprod: well after review .. I need this check since the method can take a resource or a StreamIterator and I don't want to create 2 named constructors for this otherwise you could create a CSV object from a path 👎

PR #211, opened 2017-02-01
Output File on Disk

Gabriellpweb:
Added method outputFile to save csv on disk.

### Targeted release version
dev-master

### PR Impact
It was implemented based on master. Must change $this->getCsvDocument() to $this->getIterator() to work on 8.* releases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 vendor
 build
 composer.lock
+.idea
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -1,16 +1,16 @@
 <?php
 /**
-* This file is part of the League.csv library
-*
-* @license http://opensource.org/licenses/MIT
-* @link https://github.com/thephpleague/csv/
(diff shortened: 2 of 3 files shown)


nyamsprod:
@Gabriellpweb Thanks for the PR but I don't get the reasoning for this feature ? The file is already saved to the disk when using the Writer class ? What's the added value ?

Gabriellpweb:
I need the file on disk, but the output method only provides download. So I've implemented a method to only save in the disk.

nyamsprod:
But if you use the Writer class the file *is* saved to the disk

```php
<?php

require 'vendor/autoload.php';

$csv = Writer::createFromFileObject(new SplFileObject('path/to/file.csv', 'w'));
$csv->insertAll($data_to_insert);
$csv = null;
```
if you look into `path/to/file.csv` the data is inserted. So I don't understand your PR.

Gabriellpweb:
You're right! Sorry =s! Closing!

Issue #208, opened 2017-01-30
Deprecation of League\Csv\Reader::fetch


JC5:
`League\Csv\Reader::fetch` has been marked as [deprecated](http://csv.thephpleague.com/changelog/) as per the change log but it is still used as a main example in the "[reading](http://csv.thephpleague.com/reading/)" section of the manual and there does not seem to be any indication this method is actually deprecated, nor what the alternative should be.

nyamsprod:
@JC5 yes indeed it is marked as deprecated because works as starrted on the next major release and Reader::fetch will be indeed removed from the public API in the next major release. This is the reason it is marked as deprecated. There's no alternative right now for its usage but marking it like that is a hint to users that this method will be removed. AFAIK this does not in any way prevents anyone from using the method right now.

Hope this answer your question as it is similar to #206

JC5:
It does, apologies. I am used to deprecation _after_ the new method has been presented. :wink:

joshbrw:
I came here for this exact reason. Shouldn't the method be deprecated once an alternative is available?

nyamsprod:
@joshbrw Please refer to [PHPdoc documentation](https://phpdoc.org/docs/latest/references/phpdoc/tags/deprecated.html)

joshbrw:
@nyamsprod Surely these methods are going to be superceded by some other method? The PHPDoc documentation states `If it is superceded by another method it is RECOMMENDED to add a @see tag in the same PHPDoc pointing to the new element.`, which is what I'd have expected.

nyamsprod:
> If it is superceded by another method
there's a If at the start of the sentence :)

joshbrw:
Of course. But surely you're always going to need to `fetch` data from the CSV, no?

nyamsprod:
yes and having a `@deprecated` in the docblock does not mean that the method won't work anymore.. it is an indication about where the package will go next. Really I don't see the problem with this ? This is informative only.

joshbrw:
I guess it all depends on your view of what `deprecated` means. My IDE (PHPStorm) throws up a usage warning if the method is used.

nyamsprod:
> I guess it all depends on your view of what deprecated means

Nope. The definition of `@deprecated` is the one PHPdoc put out and this is the one I'm following. I could rephrase that by saying that It depends on how PHPStorm interprets it :) .

cdekok:
[...]

PR #215, opened 2017-02-22
Bug fix #213 and #208

### LINKED ISSUE #213: fetchAssoc() fails with freshly created Reader
## Issue summary

I need to pre-process a CSV file that was given to me that contains a number of columns with duplicate names. In order to process this file I am creating a reader, writing some columns with a writer, and then getting a reader:

```php
$reader = Reader::createFromPath('path/to/file.csv');
$write = Writer::createFromStream(fopen('php://temp', 'r+'));
$columns_to_extract = [1, 2, 3, 4, 5, 43];

foreach ($reader->fetch() as $row) {
    $write->insertOne(array_intersect_key($row, $columns_to_extract));
}

$reader = $write->newReader()->setOffset(0);
```

And then I pass the `$reader` into another class that uses `fetchAssoc()`:

```php
function process(Reader $csv)
{
    foreach ($csv->fetchAssoc() as $row) {
        // do stuff with row
    }
}
```

This works fine for other CSVs, but as soon as I add this pre-processing step, I get the following error:

```
InvalidArgumentException: the specified row does not exist or is empty in importer/vendor/league/csv/src/Reader.php on line 338

Call Stack:
    0.0002     233496   1. {main}() importer/bin/import:0
    0.0255    1522728   2. League\Csv\Reader->fetchAssoc() importer/bin/import:17
    0.0255    1522928   3. League\Csv\Reader->getAssocKeys() importer/vendor/league/csv/src/Reader.php:247
    0.0255    1523208   4. League\Csv\Reader->getRow() importer/vendor/league/csv/src/Reader.php:281
```

### System informations

_(In case of a bug report Please complete the table below)_

| Information | Description |
|--------------|---------|
| League\Csv version | 8.2.0 |
| PHP/HHVM version | 5.6.29 |
| OS Platform | osx  |

## Standalone code, or other way to reproduce the problem

*See description.*

### Expected result

No error.

### Actual result

Exception thrown.

nyamsprod:
- Remove the @deprecate message from the Reader class #208
- Bug fix internal Reader::getRow when used with a StreamIterator #213


src/Modifier/StreamIterator.php
```diff
@@ -156,18 +156,17 @@ public function setFlags($flags)
      * @param array  $fields
      * @param string $delimiter
      * @param string $enclosure
-     * @param string $escape
```
shadowhand: You removed the `$escape` type hint, but the param is still available in the method.

src/Modifier/StreamIterator.php
```diff
@@ -156,18 +156,17 @@ public function setFlags($flags)
...
      *
      * @return int
      */
-    public function fputcsv(array $fields, $delimiter = null, $enclosure = null, $escape = null)
+    public function fputcsv(array $fields, $delimiter = ',', $enclosure = '"', $escape = '\\')
```
shadowhand: This changes the public interface and should not be part of a bugfix.

src/Reader.php
```diff
@@ -329,19 +317,23 @@ protected function isValidKey($value)
...
+        }
+
+        $bom_length = mb_strlen($this->getInputBOM());
+        $row[0] = mb_substr($row[0], $bom_length);
+        if ($this->enclosure == mb_substr($row[0], 0, 1) && $this->enclosure == mb_substr($row[0], -1, 1)) {
+            $row[0] = mb_substr($row[0], 1, -1);
         }
 
-        return str_getcsv($line, $this->delimiter, $this->enclosure, $this->escape);
+        return $row;
```
shadowhand: The expected return value for this method is an array, and now appears to be a string.

src/Modifier/StreamIterator.php
nyamsprod: it should be part of bugfix because it aligns the StreamIterator to how SplFileObject and fputcsv works. StreamIterator must be align to SplFileObject and it did no align.

nyamsprod: StreamIterator is marked as `@internal` so anyone using it in public API exposes himself to harm.

src/Reader.php
nyamsprod: no $row is always an array

src/Modifier/StreamIterator.php
nyamsprod: yep typo error from my part.. I'll correct that

PR #221, opened 2017-03-24
Add example for frameworks

barryvdh:
## Introduction

It is not always clear for everyone that they shouldn't just output (echo + headers) the data in frameworks. This breaks the middleware/filters etc.

## Proposal 

Add an example to the docs. We could be more specific, eg. target PSR-7 and/or Symfony, or add a Streamed Response.

Alternative could also be to provide a method the generate a PSR-7 response in the code directly.

### DIFF
--- a/docs/8.0/basic-usage.md
+++ b/docs/8.0/basic-usage.md
@@ -103,6 +103,22 @@ $reader = Reader::createFromPath('/path/to/my/file.csv');
 $reader->output("name-for-your-file.csv");
 ~~~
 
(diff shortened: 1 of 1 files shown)


nyamsprod:
following your remark.. I've added a new method [AbstractCsv::chunk](http://csv.thephpleague.com/9.0/connections/output/#outputting-the-document-into-chunks) to v9 to hopefully ease framework integration when creating a Response object. 
V9 is in review period if you have other remarks now would be the perfect time :+1:

PR #241, opened 2017-08-11
Allow remote streams in Reader

joskfg:
## Introduction

This PR will allow the Reader to use non-seekable streams like remote streams (HTTP or FTP).

To allow the reader to use that new streams, the library should add a new method to that streams type. The stream will be transformed inside the library in a seekable stream. To improve the performance, it will be create in memory is it is small (default 2MB) or in a temporary file if it is bigger. 

It is 100% backward compatible.

### Targeted release version

9.0

### PR Impact

This PR will modify the Reader API and will allow the use of remote streams out of the box.

### DIFF
--- a/tests/StreamWrapper.php
+++ b/tests/StreamWrapper.php
@@ -90,20 +90,6 @@ final class StreamWrapper
      */
     public function stream_stat()
     {
-        return [
-            'dev'     => 0,
-            'ino'     => 0,
-            'mode'    => 33206,
-            'nlink'   => 0,
-            'uid'     => 0,
-            'gid'     => 0,
-            'rdev'    => 0,
-            'size'    => 0,
-            'atime'   => 0,
-            'mtime'   => 0,
-            'ctime'   => 0,
-            'blksize' => 0,
-            'blocks'  => 0,
-        ];
+        return fstat($this->stream);
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ All Notable changes to `Csv` will be documented in this file
     - `League\Csv\Writer::getFlushThreshold`
 - Improve RFC4180 compliance
     - `League\Csv\RFC4180FieldFormatter` to format field according to RFC4180 rules
+- Add connection to non-seekable resources
+    - `League\Csv\Reader::createFromUrl` to connect non-seekable resources like HTTP.
 
 ### Deprecated
 
--- a/docs/9.0/connections/instantiation.md
+++ b/docs/9.0/connections/instantiation.md
@@ -97,3 +97,35 @@ $writer = Writer::createFromFileObject(new SplTempFileObject());
 ~~~
 
 <p class="message-warning"> The <code>SplFileObject</code> <strong>MUST</strong> be seekable otherwise a <code>RuntimeException</code> exception may be thrown.</p>
+
+## Loading from a Url (Only Reader)
+
+~~~php
+<?php
+
+public static Reader::createFromUrl(
+	string $url,
+	string $open_mode = 'r',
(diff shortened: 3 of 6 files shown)


tests/StreamWrapper.php
```diff
@@ -90,20 +90,6 @@ public function stream_seek($offset, $whence)
...
-            'gid'     => 0,
-            'rdev'    => 0,
-            'size'    => 0,
-            'atime'   => 0,
-            'mtime'   => 0,
-            'ctime'   => 0,
-            'blksize' => 0,
-            'blocks'  => 0,
-        ];
+        return fstat($this->stream);
```
joskfg: This is needed because the fake information was not right, so the stream_copy_to_stream didn't work. I could hardcode the 60 as size to pass the specific test but it could produce problems in future tests.

nyamsprod:
Hi @joskfg thanks for the submitted PR. I have some reservation if this is necessary in the library. Here's my thought.

- The library should be and is still really agnostic in how it should be used and you are suggesting adding an opinionated way on how to deals with remote streams.
- I fail to see how is this not already covered by the createFromStream named constructor added in 8.2 and still present in v9 ?
- Last but not least IMHO your solution belongs in an independent package which should then suggests using league/csv or if you can create such package in time before v9 is out the other way around. For instance, I have such factory library for instance to deal with PSR-7 but for the same reasons given above I did not included it in the main league/csv package.

joskfg:
Hi @nyamsprod, thanks forreview the PR. It tries to give a solution to the fact that PHP doesn't support seek for remote streams. This means that if you want to use the library, you need to create a seekable stream that it is local and it is not easy to do.

As you refers, it can be solved with the createFromStream. I suggested to give the functionality out of the box because the remote files usage is common. Maybe there are more solutions but this is the best I found and it is based in the solution used in makeFromString.

Finally, you are right it can be an external package because all you need is createFromStream, but the library is already implementing helping methods like createFromPath or createFromString that can be easily done with createFromStream using `createFromStream(fopen('/path/to/file.csv', 'r+'));`, so I thought that this new helper could be a good solution.

Regards

nyamsprod:
@joskfg you are absolutely right, in a ideal world league CSV should need only one constructor which accepts a (stream and a SplFileObject object. But let's be realistic, the current API covers more than 80% of the average construction possibilities. And the missing alternative are easily manageable using the current API.

The PR on itself does not add anything to the CSV parsing and manipulation. It **only** offers a new way to access a CSV document and only partially as you can not use this technique for writing to a CSV  document. So we are introducing a new layer of difficulty for the developer who will have to consume the library for a partial gain. 

If this PR is accepted on what ground will I refuse for instance a PR for handling PSR-7 objects or HttpFoundation response/request ? This is just opening a trend which IMHO will only add maintenance burden on a fairly simple package.

On the other hand providing these solutions as separate **plugins** to the main package would give a better choice to the community without having to maintain them on the main package.

My 2 cents.

joskfg:
Any idea/base about how to do a plugin?

nyamsprod:
Since you are building a Factory, you can look at this example https://github.com/bakame-php/psr7-csv-factory you could even fork it and improve it if you want

joskfg:
I won't use PSR-7 because PSR-7 is just for HTTP and this work for a lot of protocols like for example FTP (see http://php.net/manual/en/wrappers.php).

nyamsprod:
I'm not asking to use PSR7 you can use this repo as an example for adding new factory method that's all.

joskfg:
Yes, I just wanted to clarify that the PR has a wider scope :-).

Thanks

nyamsprod:
I will close this PR as you will be working on a factory package for the stream wrapper.

Issue #24, opened 2014-03-10
Allow empty enclosure and escape.


andersonamuller:
Would it be possible to allow a empty string when setting the enclosure and the escape char? A client has a special format that, for some reason, should not differentiate any chars between the delimiter. 

``` php
    public function setEnclosure($enclosure = '"')
    {
        if (1 != mb_strlen($enclosure)) {
            throw new InvalidArgumentException('The enclosure must be a single character');
        }
        $this->enclosure = $enclosure;

        return $this;
    }
```

``` php
    public function setEscape($escape = "\\")
    {
        if (1 != mb_strlen($escape)) {
            throw new InvalidArgumentException('The escape character must be a single character');
        }
        $this->escape = $escape;

        return $this;
    }
```

dakira:
@nyamsprod Hi. I'd like to re-open this bug. I need to be able to write CSV with no enclosures. I have no control over the importing app and the latter won't work with CSV files using enclosures.

While RFC 4180 exists, there is no official standard for writing CSV. Also, RFC 4180 defines the usage of enclosures as a SHOULD requirement which allows for not using them.

gocsp:
i had really trouble with no enclosure, but i found a hack (for writing): 

`
$writer->setEnclosure(chr(0))
`

nyamsprod:
@dakira I believe you can have a safe workaround using this technique

jsandeo:
It is not a requirement of the CSV standard to have a field enclosed just because there are whitespaces in it:

[See how "Grand Cherokee" is represented in this example](https://en.wikipedia.org/wiki/Comma-separated_values#Example)

Is there a way to tell a writer not to enclose a field just because it has whitespaces?

nyamsprod:
yes you could use [stream filters](http://csv.thephpleague.com/filtering/) for instance to achieve this

edgreenberg:
It should be noted that once you change the delimiter (to a TAB for instance) there is no need for enclosure around spaces or quote marks.    

Tab delimited files are pretty common, and it'd be nice to be able to write them (easily).

underdpt:
I too think that it should allow for empty enclosures, it's simpler than having to write a workaround with streamfilter.

nyamsprod:
Once again this is not a library limitation but a PHP limitation 👍 
http://php.net/manual/en/splfileobject.setcsvcontrol.php

jsandeo:
I just switched to https://github.com/parsecsv/parsecsv-for-php
[...]

PR #243, opened 2017-08-18
Enable empty enclosure, support imports into old systems

Super-Fly:
## Introduction

Enable empty enclosure

Set enclosure now accept an even empty string. Most of the old systems, for which we might create CSVs, do not accept CSV import with quote enclosure or any other, that's why the library should be flexible

## Proposal 

setEnclosure('') - removed the exception if the enclosure is less then 1 char

No need

### Targeted release version

9.0.1

### DIFF
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -354,9 +354,7 @@ abstract class AbstractCsv implements ByteSequence
      *
      * @param string $enclosure
      *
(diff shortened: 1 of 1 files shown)


nyamsprod:
Hi this issue has been raised before see #24 you can't just do that using native php functions, the only correct way to achieve empty enclosure is to use the library stream filter capabilities or to develop your own userland CSV parser

PR #245, opened 2017-08-23
Adding the enclosure stream filter

nyamsprod:
This stream filter:

- allows enforcing enclosure on all CSV field or
- removes enclosure from field with whitespace see https://bugs.php.net/bug.php?id=38301

@Super-Fly This should resolve issue #24 and PR #243. Of note, removing all enclosure is not possible because fputcsv/fgetcsv will always require the enclosure field and [RFC4180](https://tools.ietf.org/html/rfc4180#section-2) says:

> Fields containing line breaks (CRLF), double quotes, and commas  should be enclosed in double-quotes.  For example:


andre-hh:
@nyamsprod how would I use this to "remove enclosure from field with whitespace" but use them only when needed (i.e. the field contains a field delimiter)?

nyamsprod:
What you are looking for was already answered in this issue https://github.com/thephpleague/csv/issues/292 .

**WARNING: It won't work if you use CSV 9.2+ with the empty escape character**

 but if you don't care about that then that the easiest way to do it

andre-hh:
From what I see in #292 there is only an override solution for SPL file objects but not for streams, right? 

I thought that the enclosure stream filter that has been merged here would be a generic solution that could be added to any writer and would mainly only fix PHP's whitespace bug as described in https://bugs.php.net/bug.php?id=38301 ? 

For the "empty escape character" I understood that this would then break that values are enclosed that contain the delimiter as part of the string; i.e. it could break the entire CSV file then?

nyamsprod:
according to the [documentation](https://csv.thephpleague.com/9.0/interoperability/enclose-field/) 

> The EncloseField is a PHP stream filter which forces the Writer class to enclose all its record fields.

so a different filter needs to be created for your issue. I may merge your PR in the package if you find and submit an elegant solution 😉. From my point of view any descent CSV parser should have no issue with how PHP output it's CSV as long as you don't have to deal with the escape character 😄

PR #248, opened 2017-08-25
Feature/add chunk to 8x (WIP)

AC-TimRourke:
**WORK IN PROGRESS, DO NOT MERGE**

I would be interested in any immediate feedback on my proposed changes. However, I have not tested this code beyond running the unit tests. Please refrain from merging until I can confirm that this actually works. 😝 

## Introduction

This pull request backports the method `AbstractCsv::chunk` (see https://github.com/thephpleague/csv/blob/master/src/AbstractCsv.php#L277) to allow output as a streaming response in a memory-efficient way for very large CSVs.

See above.

This change is strictly additive–no incompatibility should arise.

### Targeted release version

I'm not sure how you manage releases. Please provide guidance.

### PR Impact

Should add the method `AbstractCsv::chunk` to allow memory-efficient output of large CSVs.

### DIFF
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -292,7 +292,7 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
             yield $chunk;
         }
 
-        while (!$document->eof()) {
+        while ($document->valid()) {
             yield $document->fread($length);
         }
     }
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -292,7 +292,7 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
             yield $chunk;
         }
 
-        while ($document->valid()) {
+        while (!$document->eof()) {
             yield $document->fread($length);
         }
     }
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -270,4 +270,30 @@ abstract class AbstractCsv implements JsonSerializable, IteratorAggregate
 
         return new SplFileObject($this->getStreamFilterPath(), $this->open_mode);
     }
+
+    /**
+     * Returns the CSV document as a Generator of string chunk
+     *
+     * @param int $length number of bytes read
+     *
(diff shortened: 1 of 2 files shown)


src/AbstractCsv.php
```diff
@@ -270,4 +270,30 @@ protected function setIterator()
...
+
+        $document = $this->getIterator();
+        $input_bom = $this->getInputBOM();
+        $document->rewind();
+        $document->fseek(strlen($input_bom));
+        foreach (str_split($this->output_bom.$document->fread($length), $length) as $chunk) {
+            yield $chunk;
+        }
+
+        while ($document->valid()) {
```
AC-TimRourke: @nyamsprod I just discovered that no method called `valid` exists on the object returned by `AbstractCsv::getIterator`. Is something like this an acceptable workaround?

```php
while (!$document->eof()) {
    // etc.
}
```

AC-TimRourke: Never mind, looks like `AbstractCsv::getIterator` always returns either `League\Csv\Modifier\StreamIterator`, or `SplFileObject`, both of which would allow calling `::valid` for this use case.

nyamsprod:
@AC-TimRourke thanks for the PR but I have one major objection for this as stated on the documentation website:

> Once a new major version is released, the previous stable release remains supported for six more months through patches and security fixes.

Since `AbstractCsv::chunk` is not a patch nor a security fix I can not merge your PR in the 8.x branch.

Having said that, your PR also as a bug. You are assuming that `AbstractCsv::getIterator` will always return an object which exposes the `fread` method, which is not true since 8.2 and the introduction of stream support. So What I would suggest is either fork 8.x (since you are guaranted that no new features will land on it) or use the proposed workaround.

AC-TimRourke:
👍  thanks for the heads up, I'll go ahead and close this. I'm hoping to support PHP7+ in the project I intend to use this in, so I'd prefer the 9.x branch anyway. Appreciate the feedback, and the warning about the bug!

PR #263, opened 2017-09-27
Fix possible null reference in Stream __debugInfo

Toilal:
When error occurs in constructor, before `stream` variable is defined, the error message is cryptic as __debugInfo() fails.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -147,7 +147,7 @@ class Stream implements SeekableIterator
      */
     public function __debugInfo()
     {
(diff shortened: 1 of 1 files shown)


nyamsprod:
@Toilal could you show me test code when that happens ?

Toilal:
I tried to give a SplFileObject to Writer::createFromStream() (instead of createFromFileObject).

nyamsprod:
Wouldn't that return a TypeError before even getting to `__debugInfo` ??

Issue #262, opened 2017-09-27
Argument passed must be a seekable stream resource


Toilal:
## Issue summary

I wish I could use Writer directly on `php://output` stream (Feature Request).

### System informations

| Information | Description |
|--------------|---------|
| League\Csv version | 9 |
| PHP/HHVM version | 7.1 |
| OS Platform | Linux |

## Standalone code, or other way to reproduce the problem

```php
$output = fopen('php://output', 'w');
$this->writer = Writer::createFromStream($output);
$this->writer->insertOne(['a', 'b', 'c']);
```

### Expected result

Simple use case like this one should work as expected, as it doesn't require any seek method

### Actual result

createFromStream() fails with message: `Argument passed must be a seekable stream resource`.

When commenting out the seekable check [here](https://github.com/thephpleague/csv/blob/7dfedca6b626a5ed85e7c1ffe838c8c4364ede95/src/Stream.php#L112-L114), it works as expected.

Maybe this check could be disabled and display an error message only when trying to use a feature that really requires the stream to be seekable, and [disable behavior like this](https://github.com/thephpleague/csv/blob/7dfedca6b626a5ed85e7c1ffe838c8c4364ede95/src/Writer.php#L173-L176) when stream is not seekable.

nyamsprod:
@Toilal thanks for submitting this enhancement. Don't forget to update the documentation too. As this means that a non seekable CSV can not use any of the output methods. FWIW you could have gone with 

```php
$csv = Writer::createFromFileObject(new SplFileObject('php://output', 'w'));
$csv->insertOne(['a', 'b', 'c']);
```

It works as expected but it will trigger an Warning instead of a Error if you try to use any output method. Maybe we could do something about that too ? I don't know

nyamsprod:
@Toilal just for your information I've just push a commit to enable non seekable stream. This feature is stable but will be release with version 9.1.0 ... when I have time

shameerc:
Shouldn't we also update [this exception message](https://github.com/thephpleague/csv/blob/master/src/Stream.php#L109) as it now supports non seekable streams? :)

nyamsprod:
yes it will be updated on the next patch release along with other patches in the pipe

shameerc:
Happy to make a PR, if it's not done already.

nyamsprod:
already done on my local branch but thanks for this offer 👍

PR #264, opened 2017-09-27
Add partial support for non-seekable streams (#262)

Toilal:
This is still a work in progress ... I'll update the docs later, but feel free to review this before going further.

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -44,6 +44,13 @@ class Stream implements SeekableIterator
     protected $stream;
 
     /**
+     * Tell whether the stream support seek operations
+     *
+     * @var bool
+     */
+    protected $seekable;
+
+    /**
      * Tell whether the stream should be closed on object destruction
      *
      * @var bool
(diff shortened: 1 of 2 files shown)


nyamsprod:
The more I think about it the more I'm thinking the seekable property should also be present on the AbstractCsv class too. So that the check is done once ? What do you think ?

src/Writer.php
```diff
@@ -226,6 +226,9 @@ public function addValidator(callable $validator, string $validator_name): self
      */
     public function setNewline(string $newline): self
     {
+        if ("\n" !== $this->newline && $this->document instanceof Stream && !$this->document->isSeekable()) {
```
nyamsprod: I don't think you need to add this check here... not being able to seek should not prevent setting the newline. This is to be consistent with for instance providing a non writable CSV document to the Writer class. on its own it is not forbidden but you'll get an error on insertion. It should be the same with a non seekable document.

src/Stream.php
```diff
@@ -44,6 +44,13 @@ class Stream implements SeekableIterator
     protected $stream;
 
     /**
+     * Tell whether the stream support seek operations
+     *
+     * @var bool
+     */
+    protected $seekable;
```
nyamsprod: I would prefer the property to be called is_seekable since it will hold a boolean value.

Toilal:
Do you think this PR really adds value ? It seems to add complexity without adding much value. 

Maybe adding a note in docs to tell user about [this trick](https://github.com/thephpleague/csv/issues/262#issuecomment-332543631) is enough ? I'm using it in my project and it works like a charm.

nyamsprod:
@Toilal If the trick works as expected then yes adding a note somewhere in the documentation is a good enough solution for me. If you can submit a PR with the added info in the documentation I'll gladly merge it.

Toilal:
I'll do, thanks :)

Issue #258, opened 2017-09-11
Reader::createFromPath() opens the file in write mode


Bilge:
## Issue summary

Although I realise createFromPath() is not implemented directly by Reader, the fact that this calling convention exists and yet does not open the file in read-only mode is a terrible design flaw because it's misleading and error-prone. If this method must be implemented in an abstract class, and if it must default to requiring write mode access, Reader must override the default in order for its behaviour to be sane and predictable.

### Expected behaviour

| Method | File access mode |
|-|-|
| Reader::createFromPath() | read |
| Writer::createFromPath() | write |

### Actual behaviour

| Method | File access mode |
|-|-|
| Reader::createFromPath() | read, **write** |
| Writer::createFromPath() | write |

### No justification

As shown above, the `Reader` **demands write access** despite the fact that it should not need it because it's a *reader*. Just because this fact is [poorly documented](http://csv.thephpleague.com/9.0/connections/instantiation/#loading-from-a-file-path) does not imply justification; there is no justification for this bizarre design decision in the documentation. Indeed, unless one is intimately familiar with [PHP file open nomenclature](http://php.net/manual/en/function.fopen.php), simply stating that the default is *r+* does not clearly convey that even the `Reader` class will demand write modes.

As an aside, the default should probably include the `b` flag to avoid unexpected data translation.

---
Continued from #256. Please do not close this issue prematurely.

nyamsprod:
#256 was closed because of no real issue and poor remarks from your part. If you found the documentation lacking please provide a PR to improve them or at least be constructive. Same with this issue. Instead of complaining please provide a PR as I found no issue with the current API which is documented. I don't tolerate poor remarks or attitude on open source projects that I maintain. If your don't like the current API you are still free to fork the project and do as you please but orherwise I would request that you calm down.

Bilge:
@nyamsprod You are completely missing the point and taking this too personally. This has nothing to do with documentation; it is a design flaw with the defaults you have chosen for file access modes for the `Reader` class. The default file access mode should not include the write attribute. This cannot be fixed with documentation, it can only be fixed in code.

nyamsprod:
@bilge submit a PR it will be reviewed and if it is found useful it will be merged. Thanks

Bilge:
It will be a BC break because it requires changing the default file access
mode. Are you OK with that?

nyamsprod:
Yes but it will have to wait until the next major release

Bilge:
[...]

PR #266, opened 2017-10-12
Override the createFromPath static method for a simple open mode on Reader

btmash:
## Introduction

I actually got bit by the readonly issue that is described in #258 And figured I'd give a shot at how to fix the issue.

## Proposal 

I introduce a new static method in reader where the default open mode is 'r'. I also added a test to ensure this works by adding a readonly csv file.

I override the static function ```createFromPath``` provided by AbstractCsv to set the default open mode.

There shouldn't be a BC issue.

### Targeted release version

Could be 9.0, 10.0, up to you. This is my first time writing a PR for a PHP library like this so its very much a learning experience for me 😄 

### PR Impact

I don't think there should be a PR impact.

## Open issues

#258

### DIFF
--- a/src/Reader.php
+++ b/src/Reader.php
@@ -62,6 +62,14 @@ class Reader extends AbstractCsv implements Countable, IteratorAggregate, JsonSe
     protected $stream_filter_mode = STREAM_FILTER_READ;
 
     /**
+     * @inheritdoc
+     */
+    public static function createFromPath(string $path, string $open_mode = 'r', $context = null): AbstractCsv
+    {
+        return new static(Stream::createFromPath($path, $open_mode, $context));
+    }
+
+    /**
      * Returns the header offset
      *
      * If no CSV header offset is set this method MUST return null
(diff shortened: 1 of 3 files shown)


src/Reader.php
```diff
@@ -62,6 +62,14 @@ class Reader extends AbstractCsv implements Countable, IteratorAggregate, JsonSe
     protected $stream_filter_mode = STREAM_FILTER_READ;
 
     /**
+     * @inheritdoc
+     */
+    public static function createFromPath(string $path, string $open_mode = 'r', $context = null): AbstractCsv
```
nyamsprod: This introduces a tiny BC break you can't just correct it like this

btmash: Since you'd mentioned this would potentially go in a 10.0 release (which is likely to be atleast a year away), I approached it with that in mind. Is there another way I should approach this PR?

Jalle19: The return type-hint here is wrong. You're returning `new static` so the return type should be `Reader`. While upgrading from 9.0.1 to 9.1.2 I started getting errors like these from Phan:

```
app/Legacy/Console/Commands/AbstractCsvImportCommand.php:77 PhanUndeclaredMethod Call to undeclared method \League\Csv\AbstractCsv::setHeaderOffset
```

Issue #268, opened 2017-10-14
Preventing CSV Injection


klaude:
## Issue summary

Ref: http://georgemauer.net/2017/10/07/csv-injection.html

This isn't a bug report per-se, but a temperature check to see if y'all feel that a feature to prevent injection on CSV writing would be accepted into the project. 

I'd like to add a configuration setting to new `Writer` instantiation and a parameter to `output()` to escape formulas in the CSV document. If the user opts to escape formulas then any cell in the CSV that begins with the "=", "-", "+", or "@" characters will have a `\t` prepended to it. There are valid use cases for executable code in CSV documents, so injection prevention would default to false. 

I'd target the latest major version of the project, but don't mind back-porting this to version 8. I'm willing to do all the work, but only if this is something y'all feel is important. If y'all want to see this in the project, but feel there's a better way to have the user configure injection prevention I'm all ears. Let me know what you think. Thanks!

nyamsprod:
@klaude For what I understand about csv injection, you do not need to add anything to league csv to make it work. as we talk the package already allow adding some sanity/escaping check prior to inserting your data to the CSV document. So what you really need is provide a clean implementation of csv injection prevention in form of a [Formatter](http://csv.thephpleague.com/9.0/writer/#records-filtering) and/or a [StreamFilter](http://csv.thephpleague.com/9.0/connections/filters/). Depending on your work I may found it useful, maintainable and thus incorporate it but I would reserve my final thought upon seeing the PR... not before.

klaude:
That's a good call, @nyamsprod. It makes a lot more sense than a parameter bag. I'll come up with something and get you a PR. Thanks!

nyamsprod:
@klaude having read in more depth the article you linked I have to wonder if this is a job to be shipped with League CSV. Basically the article stress that this is not a CSV issue per se but a Spreadsheet one and instead of fixing spreadsheet software we are about to pseudo fix a CSV document without the knowledge of how and who is going to use it. That a pretty dangerous decision to make ... In some regards this reminds me of PHP's escaping data model which was broken to begin with

klaude:
I agree. That's why it makes sense to leave injection handling to the user's discretion. I feel the library should provide it so users get an easier security option. 

[...]

PR #269, opened 2017-10-16
Adding EscapeFormulaInjection formatter

nyamsprod:
view #268


nyamsprod:
@klaude here's a POC which works so it was simple to add But I'm still skeptical in its usefulness and/or binding it to the package

klaude:
That looks good to me. I think the only difference between this and what I was envisioning was a constructor argument to say which columns to escape, but this suits my purposes at work just fine.

PR #275, opened 2017-11-12
Fix if using a custom set_error_handler() causing error instead of an…

### LINKED ISSUE #254: Invalid error response for non-existent file
## Issue summary

Attempting to load a non-existent file throws a fatal error instead of a League\Csv\Exception.

### System informations

| Information | Description |
|--------------|---------|
| League\Csv version | 9.0.1 |
| PHP/HHVM version | 7.0.22-0ubuntu0.16.04.1 |
| OS Platform | ubuntu |

## Standalone code, or other way to reproduce the problem

`Reader::createFromPath('/does/not/exist.csv');`

### Expected result
Similar to Flysystem
```
[League\Flysystem\FileNotFoundException]
File not found at path: /does/not/exist.csv
```

### Actual result
```
[Symfony\Component\Debug\Exception\FatalThrowableError]
Wrong parameters for League\Csv\Exception([string $message [, long $code [, Throwable $previous = NULL]]])
```

csiszarattila:
Fix #254

### DIFF
--- a/src/Stream.php
+++ b/src/Stream.php
@@ -176,7 +176,11 @@ class Stream implements SeekableIterator
         }
 
         if (!$resource = @fopen(...$args)) {
-            throw new Exception(error_get_last()['message']);
+            throw new Exception(
+                error_get_last()
+                ? error_get_last()['message']
+                : 'failed to open stream: No such file or directory'
+            );
         }
 
         $instance = new static($resource);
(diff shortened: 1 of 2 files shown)


nyamsprod:
@csiszarattila I've updated your PR by using PHP7 null coalescence operator. There was another occurrence of error_get_last in the codebase so I've applied your fix their too and improved the test suite. thanks for the contribution. 👍

PR #283, opened 2018-02-20
Static analysis with PHPStan

lookyman:
Hmm, I have no idea why Scrutinizer fails..

### DIFF
--- a/src/Writer.php
+++ b/src/Writer.php
@@ -115,7 +115,8 @@ class Writer extends AbstractCsv
     /**
      * Adds a single record to a CSV document
      *
-     * @param string[] $record an array
+     * @param array $record An array consisting of nulls or values that can be converted to string (scalar string, int,
+     *                      float, objects implementing __toString() method.
      *
      * @throws CannotInsertRecord If the record can not be inserted
      *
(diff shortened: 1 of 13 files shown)


src/Writer.php
```diff
@@ -126,7 +126,7 @@ public function insertOne(array $record): int
         $record = array_reduce($this->formatters, [$this, 'formatRecord'], $record);
         $this->validateRecord($record);
         $bytes = $this->document->fputcsv($record, $this->delimiter, $this->enclosure, $this->escape);
-        if (!$bytes) {
+        if (!(bool) $bytes) {
```
nyamsprod: I'd rather write 
```php
<?php

if (false === $bytes) {
}
```
since `fputcsv` returns `false` on error but may also return `0`

src/Writer.php
```diff
@@ -115,7 +115,7 @@ public function insertAll($records): int
     /**
      * Adds a single record to a CSV document
      *
-     * @param string[] $record an array
+     * @param (string|int|null)[] $record an array
```
nyamsprod: while I understand this notation I believe it is as incomplete as the one you are trying to fix for instance does your notation covers this usage ?

```php
$csv = Writer::createFromString('');
$csv->insertOne([
    new class {
        function __toString() {
            return 'foo';
        }
    },
    'bar',
]);

echo $csv->getContent();
```

lookyman: Hmm maybe it should just be `mixed[]` then?

src/Writer.php
lookyman: Will do.

src/Writer.php
nyamsprod: I'd rather use `array` and specify in the doc bloc that flat array consisting of scalar null and object which implements __toString values can be used IMHO mixed is just too permissive

lookyman: I don't think that is possible to express in the typehint though..

nyamsprod: no it's not that why I'm leaning toward just using the default typehint which is `array` and explain in the docblock exactly what is expected

lookyman: Right, I got ya.

lookyman:
Ok, I fixed the issues. I can squash the commits if you want..

nyamsprod:
yes please squash them 👍

nyamsprod:
thanks for the PR 👍 🎉

PR #285, opened 2018-03-03
Fix incorrectly overriden static type-hint in Reader and AbstractCsv

Jalle19:
## Introduction

The following code produces this Phan error: `PhanUndeclaredMethod Call to undeclared method \League\Csv\AbstractCsv::setHeaderOffset`:

```php
$csvReader = Reader::createFromPath((string)$this->argument('csvPath'));
$csvReader->setHeaderOffset(0);
```

## Proposal 

Remove the static return type-hint from `AbstractCsv` since sub-classes now override `createFromPath()`.

None

### Targeted release version

9.1.2

## Open issues

There's some references to this in https://github.com/thephpleague/csv/pull/266

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ composer.lock
 docs/_site
 vendor
 /nbproject/private/
+.php_cs.cache
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -155,7 +155,7 @@ abstract class AbstractCsv implements ByteSequence
      *
      * @return static
      */
-    public static function createFromPath(string $path, string $open_mode = 'r+', $context = null): self
+    public static function createFromPath(string $path, string $open_mode = 'r+', $context = null)
     {
         return new static(Stream::createFromPath($path, $open_mode, $context));
     }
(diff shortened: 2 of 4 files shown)


nyamsprod:
@Jalle19 thanks for your PR. Indeed when 9.0.0 was released this bug on the return type of all the named constructors methods was introduced. The issue that I have is that your resolution may be  considered a BC break by some. If it's not a BC break then you also need to fix the other named constructor return type (ie all the `createFrom*` methods) ? 
Bottom line, if you can convince me that this is not a BC break and if you complete the PR by fixing the other named constructors I'll gladly merge your PR.

Jalle19:
I guess I'll be fixing the other createFrom methods too then. I have a hard time coming up with a scenario where this would break anything, if I do I can mention it here.

nyamsprod:
@Jalle19 TBH I too don't think it is BC break since removing the return type does not restrict or change the current public API. So IMHO we can move on with your PR once finalized ... unless someone disagree  :+1:

Jalle19:
@nyamsprod I removed all the relevant static return type-hints

nyamsprod:
let's merge this .. I'll release a new CSV patch version later in the week if I forgot just ping me here again 👍

Jalle19:
Thanks!

PR #309, opened 2018-09-07
Adding support for the escape char with an empty string via polyfill

nyamsprod:


nyamsprod:
@theodorejb I've added a new class `RFC4180Iterator` in the PR to polyfill the behavior with the Reader class it was easier than I thought anyway. Could you give it a spin .. I've try to keep the behavior as close to fgetcsv as I could.

theodorejb:
I tested the branch again and did some benchmarking. The good news is that it seems to be working reliably! I was able to output a large CSV, read it into memory, and write it back again with zero loss of data.

I compared the performance of native reading/writing (passing a null byte to `setEscape`) with the polyfills (passing a blank string) on an 11MB, ~24K row CSV.  Here are the results (averaged over 5 runs):

| Action                   | Average time (sec) | Percent slower |
| ------------------------ | ------------------ | -------------- |
| native read native write | 4.552              | N/A            |
| native read custom write | 5.008              | 10%            |
| custom read native write | 7.496              | 65%            |
| custom read custom write | 7.960              | 75%            |

While the custom writer is almost as fast as the native function, the custom reader is noticeably slower. 😢 Correct results are more important than performance, though, and maybe the custom reader can be optimized more in the future.

From reviewing the patch diff, it looks like the custom reader unit tests are just checking that the number of rows is correct. Shouldn't they also verify that the row contents match what is expected?

nyamsprod:
@theodorejb indeed the test suite should be improved be I kept improving the parsing using local data. I will update them. This is a first implementation and it uses fgetc which is slow but accurate maybe I can improve it by switching to fgets but I wanted first to have an accurate code before optimization

theodorejb:
It doesn't seem that `fgets` is any faster - maybe even a bit slower. With the latest commit it now takes an average of 8.6 seconds to parse and output my CSV.

theodorejb:
@nyamsprod For reference here's the code I used for testing:
```php
use League\Csv\{Reader, Writer};

require 'vendor/autoload.php';

$start = microtime(true);
$csv = Reader::createFromPath('test.csv');
$csv->setEscape("");

$copy = Writer::createFromPath('test_copy.csv', 'w');
$copy->setEscape("");
$copy->insertAll($csv->getRecords());

echo 'took ' . (microtime(true) - $start) . ' sec';
```

I verified that test_copy.csv is byte-for-byte identical to the original file by checking its SHA-256 hash. I'm sure there are edge cases in reading CSVs that my file doesn't cover, though, so hopefully the League/CSV unit tests will catch all of them.

nyamsprod:
@theodorejb understood you had some chances with your file then :smile:  because the poly fill mimics `fputcsv` in that it add enclosure if an space character is found in a field content. This means that your CSV does not contains spaces or that they are already quoted. Using your test script with a CSV containing more than 50000 records, a simple `md5_file` returns false because of that but a check record by records validate that the documents contains the same data. :+1: . I'll let this PR as it is right now and merge it to the master branch hopefully next week.

PR #318, opened 2018-11-29
Fixes spelling in exception

lordrhodos:
Fixes misspelled `equel` in exception message of `\League\Csv\Statement::limit`


nyamsprod:
Good catch thanks 😉👍

PR #329, opened 2019-02-21
Add php 7.4 build to travis

chalasr:


chalasr:
PHPCSFixer failure fixed, this should be ready.

PR #332, opened 2019-02-26
Update Stream.php

HaozhouChen:
Fix the seek method in League\Csv\Stream.php. If the $position parameter is 0, $offset will become -1.


nyamsprod:
could you:
- add the tests in StreamTest.php and then I'll merge it 
- use Yoda style in the if condition 
and I'll merge your fix

thanks in advance

PR #337, opened 2019-03-29
Update ResultSet.php

netbull:
actually return the json encoded


nyamsprod:
What's the reasoning behind this PR? Not only is it BC breaking but also I fail to understand it's need 🤔

netbull:
was the method suppose to return a json encoded string as the method name states? now it returns plain array.. or probably i'm missing something :/

nyamsprod:
Indeed you're missing the fact that the class is correctly implementing a PHP interface 😉

netbull:
i'm such an idiot... sorry to bother..

PR #345, opened 2019-06-14
Fix phpdoc

ghobaty:
_The things I do to get my IDE to stop complaining..._

### DIFF
--- a/src/Reader.php
+++ b/src/Reader.php
@@ -43,8 +43,8 @@ use const STREAM_FILTER_READ;
  * A class to parse and read records from a CSV document.
  *
  * @method array fetchOne(int $nth_record = 0) Returns a single record from the CSV
- * @method Generator fetchColumn(string|int $column_index) Returns the next value from a single CSV record field
- * @method Generator fetchPairs(string|int $offset_index = 0, string|int $value_index = 1) Fetches the next key-value pairs from the CSV document
+ * @method \Generator fetchColumn(string|int $column_index) Returns the next value from a single CSV record field
+ * @method \Generator fetchPairs(string|int $offset_index = 0, string|int $value_index = 1) Fetches the next key-value pairs from the CSV document
  */
 class Reader extends AbstractCsv implements Countable, IteratorAggregate, JsonSerializable
 {


nyamsprod:
Nice catch I prefer declaring the import on top of the class. Once this is done I'll gladly merge your PR. Thanks for the contribution

cafferata:
>*Nice catch I prefer declaring the import on top of the class.*

@nyamsprod, the import is already on top of the class.
*https://github.com/thephpleague/csv/blob/master/src/Reader.php#L19*

@ghobaty, when you are into fixing the DocBlocks, there are more missing/wrong DocBlocks in de codebase (11x). @ghobaty would you like to fix this? Or shall I do this?

```diff
    /**
     * Determine the CSV record header.
     *
     * @param int $offset
     *
+    * @return string[]
     * @throws Exception If the header offset is set and no record is found or is the empty array
     */
    protected function setHeader(int $offset): array
    {
```

```diff
    /**
     * Returns the row at a given offset.
     *
+    * @param int $offset
     *
     * @return array|false
     */
    protected function seekRow(int $offset)
    {
```

ghobaty:
@cafferata please feel free to!

nyamsprod:
@cafferata there are not missing. If the argument is defined using PHP7 parameter type hinting they are not reported again using docblock. It enables better maintenance :wink

cafferata:
@nyamsprod I partially agree with you. Yes it can be a choice to omit the DocBlocks in connection with the PHP 7 parameter type hinting. However, the existing DocBlocks have multiple `@param` hints that are not always complete/correct.

Examples:

```diff
   /**
     * Retuns the CSV document as a Generator of string chunk.
     *
-    * @param int $length number of bytes read
-    *
     * @throws Exception if the number of bytes is lesser than 1
     */
    public function chunk(int $length): Generator
    {
```
*\League\Csv\AbstractCsv::chunk*

```diff
   /**
     * New instance.
     *
-    * @param string   $escape        escape character to escape each CSV formula field
-    * @param string[] $special_chars additional spreadsheet formula starting characters
-    *
     */
    public function __construct(string $escape = "\t", array $special_chars = [])
    {
```
*\League\Csv\EscapeFormula::__construct*

In addition, there is a small problem in the example above. The default value for parameters with a class type can only be `NULL`.

```json
php: >=7.0.10
```
*https://github.com/thephpleague/csv/blob/master/composer.json#L24*

nyamsprod:
@cafferata maybe create a separate thread issue for you suggestion as this current PR was already accepted and I prefer having one issue per thread if possible. you can even propose a PR and we can discuss it there .

Issue #344, opened 2019-06-12
HTML output table header row


adrienne:
_(Fill in the relevant information below to help triage your issue.)_

|    Q        |   A
|------------ | ------
| Version     | 9.2

### Question

I have the following code, expecting that my HTML conversion would output the header row with `<th>` tags. Instead it doesn't output the first row at all. 

```
$reader = Reader::createFromString($csvString);
$reader->setHeaderOffset(0);
$converter = (new HTMLConverter())
        ->table('dimensions-table')
        ->tr('data-record-offset')
        ->td('title');

$html = $converter->convert($reader);
```

If I remove `$reader->setHeaderOffset(0);` it does output the first row, but there is still no table header with `<th>` elements. Is there a way to properly output a header row?

nyamsprod:
@adrienne thanks for using the library. Currently there's no way to assign the `<th>` tag to the converted HTML. But I'll welcome any PR adding this functionality to the current codebase. Should not even BC break.

nyamsprod:
@adrienne to be more specific this class converts any tabular data not just the reader so the main issue if one wants to add this feature is to figure out how to make it works with any tabular data not just the Reader object.

PR #348, opened 2019-07-11
Allow importing the generated XML in a DOMDocument and improving  HTML converter output

kusabi:
In reference to #344 

Added an optional parameter to the `HTMLConverter::convert($records, $headers = null)` method.

When an array of headers is passed, the HTML table will contain a row of `<th>` elements.

The code in question goes from...

```php
$reader = Reader::createFromString($csvString);
$reader->setHeaderOffset(0);
$converter = (new HTMLConverter())
        ->table('dimensions-table')
        ->tr('data-record-offset')
        ->td('title');

$html = $converter->convert($reader);
```
and becomes...

```php
$reader = Reader::createFromString($csvString);
$reader->setHeaderOffset(0);
$converter = (new HTMLConverter())
        ->table('dimensions-table')
        ->tr('data-record-offset')
        ->td('title');

$html = $converter->convert($reader, $reader->getHeader());
```

### DIFF
--- a/src/HTMLConverter.php
+++ b/src/HTMLConverter.php
@@ -58,19 +58,45 @@ class HTMLConverter
     /**
      * Convert an Record collection into a DOMDocument.
      *
-     * @param array|Traversable $records the tabular data collection
+     * @param array|Traversable $records      The tabular data collection
+     * @param array             $headerRecord An optional array of headers to output to the table using `<thead>` and `<th>` elements
+     * @param array             $footerRecord An optional array of footers to output to the table using `<tfoot>` and `<th>` elements
+     *
+     * @return string
      */
-    public function convert($records): string
+    public function convert($records, array $headerRecord = [], array $footerRecord = []): string
     {
-        /** @var DOMDocument $doc */
-        $doc = $this->xml_converter->convert($records);
+        if ([] === $headerRecord && [] === $footerRecord) {
+            /** @var DOMDocument $doc */
+            $doc = $this->xml_converter->convert($records);
 
-        /** @var DOMElement $table */
-        $table = $doc->getElementsByTagName('table')->item(0);
-        $table->setAttribute('class', $this->class_name);
-        $table->setAttribute('id', $this->id_value);
+            /** @var DOMElement $table */
+            $table = $doc->getElementsByTagName('table')->item(0);
+            $this->styleTableElement($table);
 
-        return $doc->saveHTML($table);
+            return $doc->saveHTML($table);
+        };
+
+        $doc = new DOMDocument('1.0');
+
+        $tbody = $this->xml_converter->rootElement('tbody')->import($records, $doc);
+        $table = $doc->createElement('table');
+        $this->styleTableElement($table);
+        if (!empty($headerRecord)) {
+            $table->appendChild(
+                $this->createRecordRow('thead', 'th', $headerRecord, $doc)
+            );
+        }
+        $table->appendChild($tbody);
+        if (!empty($footerRecord)) {
+            $table->appendChild(
+                $this->createRecordRow('tfoot', 'th', $footerRecord, $doc)
+            );
+        }
+
+        $doc->appendChild($table);
+
+        return $doc->saveHTML();
     }
 
     /**
@@ -110,5 +136,38 @@ class HTMLConverter
         $clone->xml_converter = $this->xml_converter->fieldElement('td', $fieldname_attribute_name);
 
         return $clone;
+    }
+
+    /**
+     * Create a DOMElement representing a single record of data
+     *
+     * @param string $recordTagName
+     * @param string $fieldTagName
+     * @param array $record
+     * @param DOMDocument $doc
+     *
+     * @return DOMElement
+     */
+    private function createRecordRow(string $recordTagName, string $fieldTagName, array $record, DOMDocument $doc) : DOMElement
+    {
+        $node = $this->xml_converter->rootElement($recordTagName)->fieldElement($fieldTagName)->import([$record], $doc);
+
+        /** @var DOMElement $element */
+        foreach ($node->getElementsByTagName($fieldTagName) as $element) {
+            $element->setAttribute('scope', 'col');
+        }
+
+        return $node;
+    }
+
+    /**
+     * Style the table dom element
+     *
+     * @param DOMElement $table_element
+     */
+    private function styleTableElement(DOMElement $table_element)
+    {
+        $table_element->setAttribute('class', $this->class_name);
+        $table_element->setAttribute('id', $this->id_value);
     }
 }
--- a/src/XMLConverter.php
+++ b/src/XMLConverter.php
@@ -80,11 +80,34 @@ class XMLConverter
     ];
 
     /**
-     * Convert an Record collection into a DOMDocument.
+     * Convert a Record collection into a DOMDocument.
      *
      * @param array|Traversable $records the CSV records collection
+     *
+     * @return DOMDocument
      */
     public function convert($records): DOMDocument
+    {
+        if (!is_iterable($records)) {
+            throw new TypeError(sprintf('%s() expects argument passed to be iterable, %s given', __METHOD__, gettype($records)));
+        }
+        $doc = new DOMDocument('1.0');
+        $doc->appendChild(
+            $this->import($records, $doc)
+        );
+        return $doc;
+    }
+
+    /**
+     * Create a new DOMElement related to the given DOMDocument.
+     *
+     * **DOES NOT** attach to the DOMDocument
+     *
+     * @param iterable $records
+     *
+     * @return DOMElement
+     */
+    public function import($records, DOMDocument $doc): DOMElement
     {
         if (!is_iterable($records)) {
             throw new TypeError(sprintf('%s() expects argument passed to be iterable, %s given', __METHOD__, gettype($records)));
@@ -92,15 +115,13 @@ class XMLConverter
 
         $field_encoder = $this->encoder['field']['' !== $this->column_attr];
         $record_encoder = $this->encoder['record']['' !== $this->offset_attr];
-        $doc = new DOMDocument('1.0');
         $root = $doc->createElement($this->root_name);
         foreach ($records as $offset => $record) {
             $node = $this->$record_encoder($doc, $record, $field_encoder, $offset);
             $root->appendChild($node);
         }
-        $doc->appendChild($root);
 
-        return $doc;
+        return $root;
     }
 
     /**
--- a/tests/HTMLConverterTest.php
+++ b/tests/HTMLConverterTest.php
@@ -54,6 +54,59 @@ class HTMLConverterTest extends TestCase
         self::assertContains('<table class="table-csv-data" id="test">', $html);
         self::assertContains('<tr data-record-offset="', $html);
         self::assertContains('<td title="', $html);
+        self::assertNotContains('<thead>', $html);
+        self::assertNotContains('<tbody>', $html);
+        self::assertNotContains('<tfoot>', $html);
+    }
+
+    /**
+     * @covers ::__construct
+     * @covers ::table
+     * @covers ::tr
+     * @covers ::td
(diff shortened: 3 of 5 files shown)


nyamsprod:
@kusabi thanks for the PR I see a great deal of thoughts has been put into it so really I appreciate it.

I have some reservations on some UX decision made:

- The first one is the use of `boolean` value. I'm don't like boolean values because they are boolean and can be confusing at best.
- I do not want to change how XMLConverter works. As it is this class does exactly what it is supposed to do and any XML client can use the returned XML without any issue so we should not changed this class.
- Last but not least,  whatever we do we won't have people liking the solution 100% so to me we should catered to the 80% and leave the rest to the more experienced ones who will surely figure out a way around the limitation and basically rebuild they own HTMLConverter based on the `XMLConverter`. 

So what I would proposed instead and in line with my remarks on #344  is the following public API changed:

```php
<?php

HTMLConverter::convert(iterable $records, array $headerRecord = []): string
```

On usage it would look like this

```php
<?php

$reader = Reader::createFromString($csvString);
$reader->setHeaderOffset(0);
$converter = (new HTMLConverter())
    ->table('table-csv-data', 'users')
    ->tr('data-record-offset')
    ->td('title')
;
$html = $converter->convert($reader, $reader->getHeader());
```

And if `$headerRecord` is not empty the resulting table would look like this 

```html
<table class="table-csv-data" id="users">
<thead>
<tr>
<th scope="col">firstname</td>
<th scope="col">lastname</td>
<th scope="col">email</td>
</tr>
</thead>
<tbody>
<tr data-record-offset="0">
<td title="firstname">john</td>
<td title="lastname">doe</td>
<td title="email">john.doe@example.com</td>
</tr>
<tr data-record-offset="1">
<td title="firstname">jane</td>
<td title="lastname">doe</td>
<td title="email">jane.doe@example.com</td>
</tr>
</tbody>
<tfoot>
<tr>
<th scope="col">firstname</td>
<th scope="col">lastname</td>
<th scope="col">email</td>
</tr>
</tfoot>
```

What do you think ?

kusabi:
That's sounds fine, I will look into a solution after work.

My only concern with this approach is that without any modification to the XML converter, the HTML converter is going to have to get a little hacky.

If you are OK with this I will look into your proposed approach.

kusabi:
Ok. So attempt number 2. It is a little more concise this way and not nearly as hacky as I though it was going to get. 

What do you think?

nyamsprod:
@kusabi I took the liberty to research the HTML5 and DOM spec and also what is capable or not in PHP regarding what we are trying to do. Here's what I did find:

- from HTML spec if you add a THEAD section you are required to add a TBODY one (which we don't in your current implementation)
- from DOM spec there's no simple way to change the parent node name without some cumbersome technique (limitation from DOM)
- the XMLConverter does a great job in checking and validating the generated XML/HTML and we should let it continue doing that (ease of maintainance).

So I came to this conclusion that a better implementation would be to add **one public method** to `XMLConverter`.

```php
XMLConverter::import(iterable $records, DOMDocument $doc): DOMElement;
```

The import method creates a `DOMElement` related to the given `DOMDocument` but **does not attach it to it**. The great part of this addition is that this method can be extracted from the current `XMLConverter::convert` implementation with ease. (ie: we do not need to re-invent the wheel)

Once this method is extracted/implemented then adding the new functionnality to the `HTMLConverter` class is a breeze.

```php
HTMLConverter::convert(iterable $records, array $headerRecord = []): string
```

And here's a pseudo code

```php
public function convert(iterable $records, array $headerRecord = [])
{
    if ([] === $headerRecords) {
        // current implementation does not change
    }

    $doc = new DOMDocument('1.0');

    $thead = $this->createHeaders('thead', $headerRecord, $doc);
    $tfoot = $this->createHeaders('tfoot', $headerRecord, $doc);
    $tbody = $this->xml_converter->rootElement('tbody')->import($records, $doc);

    $table = $doc->createElement('table');
    $table->setAttribute('class', $this->class_name);
    $table->setAttribute('id', $this->id_value);
    $table->appendChild($thead);
    $table->appendChild($tfoot);
    $table->appendChild($tbody);

    $doc->appendChild($table);

    return $doc->saveHTML($table);
}

private function createHeaders(array $record, string $tagName, DOMDocument $doc)
{
    $node = $this->xml_converter->rootElement($tagName)->fieldElement('th')->import([$record], $doc);
    foreach ($node->getElementsByTagName('th') as $th) {
        $th->setAttribute('scope', 'col');
    }

    return $node;
}
```

IMHO this is much easier to understand and to implement and we open the door for those who want a different implementation to do it as they wish using the XMLConverter` object.

kusabi:
No problem, switched to your proposed solution.

If any headers or footers are added then the table has a `<thead>`, `<tbody>`, `<tfoot>`.

I also made the footer and headers optional to each other (have one without the other).

Let me know what you think.

src/HTMLConverter.php
```diff
@@ -58,19 +58,42 @@ public function __construct()
...
+
+        $doc = new DOMDocument('1.0');
+
+        $thead = $this->createRecordRow('thead', 'th', $headerRecord, $doc);
+        $tfoot = $this->createRecordRow('tfoot', 'th', $footerRecord, $doc);
+        $tbody = $this->xml_converter->rootElement('tbody')->import($records, $doc);
+        $table = $doc->createElement('table');
+        $table = $this->styleTableElement($table);
+
+        $table = $this->styleTableElement($table);
```
nyamsprod: added twice 😄

src/HTMLConverter.php
```diff
@@ -111,4 +134,40 @@ public function td(string $fieldname_attribute_name): self
...
+     * Create a DOMElement representing a single record of data
+     *
+     * @param string $rowTagName
+     * @param string $elementTagName
+     * @param array $record
+     * @param DOMDocument $doc
+     *
+     * @return DOMElement
+     */
+    private function createRecordRow(string $rowTagName, string $elementTagName, array $record, DOMDocument $doc)
```
nyamsprod: you can typehint the return type directly thanks PHP7 😉

src/HTMLConverter.php
```diff
@@ -58,19 +58,42 @@ public function __construct()
...
+     * @param array             $footerRecord An optional array of footers to output to the table using `<tfoot>` and `<th>` elements
+     *
+     * @return string
      */
-    public function convert($records): string
+    public function convert($records, array $headerRecord = [], array $footerRecord = []): string
     {
-        /** @var DOMDocument $doc */
-        $doc = $this->xml_converter->convert($records);
+        // If we do not want to add any headers or records, then we can default entirely to the XML converter
```
nyamsprod: This comment is unnecessary the code is self explaining in itself

kusabi: removed

src/HTMLConverter.php
kusabi: added the PHP7 hint but not removed the phpdock hint as I was not sure if you implied that or not?

src/HTMLConverter.php
kusabi: doh! removed

src/HTMLConverter.php
```diff
@@ -111,4 +132,40 @@ public function td(string $fieldname_attribute_name): self
...
+    }
+
+    /**
+     * Style the table dom element
+     *
+     * @param DOMElement $table_element
+     *
+     * @return DOMElement
+     */
+    private function styleTableElement(DOMElement $table_element)
```
nyamsprod: return typed whenever you can

tests/HTMLConverterTest.php
```diff
@@ -54,6 +54,63 @@ public function testToHTML()
...
+
+        $records = $stmt->process($csv);
+
+        $converter = (new HTMLConverter())
+            ->table('table-csv-data', 'test')
+            ->td('title')
+            ->tr('data-record-offset')
+        ;
+
+        $html = $converter->convert($records, $records->getHeader());
```
nyamsprod: if you only provide a `thead` why does the `tfoot` is present 🤔

tests/XMLConverterTest.php
```diff
@@ -90,4 +91,45 @@ public function testXmlElementTriggersTypeError()
...
+        $converter = (new XMLConverter())
+            ->rootElement('csv')
+            ->recordElement('record', 'offset')
+            ->fieldElement('field', 'name')
+        ;
+
+        $doc = new DOMDocument('1.0');
+        $element = $converter->import($records, $doc);
+
+        self::assertInstanceOf(DOMDocument::class, $doc);
```
nyamsprod: 👍

src/HTMLConverter.php
kusabi: Updated

tests/HTMLConverterTest.php
kusabi: Honestly, I assumed that they came as a threesome...
I originally hid them accordingly, would you you prefer I did that again?

src/HTMLConverter.php
```diff
@@ -111,4 +132,40 @@ public function td(string $fieldname_attribute_name): self
...
+     * Create a DOMElement representing a single record of data
+     *
+     * @param string $rowTagName
+     * @param string $elementTagName
+     * @param array $record
+     * @param DOMDocument $doc
+     *
+     * @return DOMElement
+     */
+    private function createRecordRow(string $rowTagName, string $elementTagName, array $record, DOMDocument $doc) : DOMElement
```
nyamsprod: rename 
- `$rowTagName` to `$recordTagName`
- `$elementTagName ` to `$fieldTagName`

for consistency in naming with the reste of the package.

kusabi: renamed

src/XMLConverter.php
```diff
@@ -80,27 +80,45 @@ class XMLConverter
...
+     *
+     * **DOES NOT** attach to the DOMDocument
+     *
+     * @param iterable $records
+     *
+     * @return DOMElement
+     */
+    public function import($records, DOMDocument $doc): DOMElement
     {
         if (!is_iterable($records)) {
```
nyamsprod: I'm wondering if we should not duplicate this check in the `XMLConverter:convert` method otherwise we are introducing a subtle BC break when throwing exception 🤔  With the current implementation the exception will always comes from the import which is true but not expected from a BC point of view .

tests/HTMLConverterTest.php
nyamsprod: yes hides them accordingly ... I prefer this to be explicit. I was unable to find a resource stating for instance that if you use `tfoot` then `thead` is required. The only requirement is that if one is present then `tbody` **must be used**.

src/XMLConverter.php
kusabi: Agree. Added to both methods.

src/HTMLConverter.php
```diff
@@ -111,4 +132,40 @@ public function td(string $fieldname_attribute_name): self
...
+    }
+
+    /**
+     * Style the table dom element
+     *
+     * @param DOMElement $table_element
+     *
+     * @return DOMElement
+     */
+    private function styleTableElement(DOMElement $table_element) : DOMElement
```
nyamsprod: I don't think `styleTableElement` needs to return anything the object is already known. :wink:

tests/HTMLConverterTest.php
kusabi: done

src/HTMLConverter.php
kusabi: changed

nyamsprod:
@kusabi here's the changes I made to your code after the PR was merged:

- made the methods protected and not private because if someone extends the class (which is possible currently it may fails for him - the class should be made final in the next major release to avoid that)
- I've changed the protected methods and parameters names to be more inline with the rest of the package (parameter should be snake_cased ...)
- Completely replace the use of `XMLConverter::convert` with `XMLConverter::import` in the `HTMLConverter` class.

I hope I'll be able to release the new version `9.3.0` as soon as possible.

Issue #352, opened 2019-08-19
BOM detection restricts stream usage with the Reader class


clcastro87:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

#### Proposal

_I would like to be able to use I/O streams. It's a great feature to support reading and writing from/to streams. But actually I/O streams are not supported (php://stdin, php://stdout)_

nyamsprod:
@clcastro87 could you be more specific as the library supports PHP stream since version 8.2 [see documentation](https://csv.thephpleague.com/9.0/connections/instantiation/) with the Reader supporting any input stream and the writer writing to any output stream

clcastro87:
@nyamsprod I'm using version 9.3. The issue is only for php://stdin. When I'm trying to read from stdin is throwing `stream does not support seeking`. I suppose is because php://stdin is not seekable. On file src/Stream:483 there is a check for this. My use case is that I'm trying to pipe to the php process in input using the console. Ex: cat myfile.csv | grep important | php mytest.php. I know maybe I can copy to a tmpfile and then process the stream but for huge file this is an extra waiting because I'll need to copy the entire input stream to a temp file and then start processing. I hope that you can understand my case. Thanks!

nyamsprod:
@clcastro87 I think this exception is only triggered if you try to setup the header of your CSV file ... so yep if you really need the header it's obvious that the CSV needs to be seekable ... so if you don't need the header it should be fine IMHO ... I haven't tested it so I may be wrong.

clcastro87:
@nyamsprod Sorry for the delay but I've been very busy the last two days. As you said of course if I try to read from stdin I'll need the field names (header). Maybe I'm doing something dumb but here you can have a simple piece of code to prove what I\m saying. 
`
$res = fopen('php://stdin', 'r');
$fields = [
    "title", 
    "brand", 
    "link", 
    "image_link",
    "price", 
    "description", 
    "gender", 
    "id", 
    "availability", 
    "google_product_category", 
    "age_group", 
    "size", 
    "color", 
    "adult", 
    "sale_price", 
    "gtin", 
    "mpn"
];
$reader = League\Csv\Reader::createFromStream($res);
foreach ($reader->getRecords($fields) as $item) {
    var_dump($item);
}
`
The error is the following:
`
carlos@carlos-laptop:~/ethyl$ tail test/Products.csv | php src/test2.php 
[...]

PR #353, opened 2019-08-23
Adding the ability to control skipping the document input BOM

nyamsprod:
This PR resolves issue #352 by re-introducing a way to control if BOM stripping should be done. This helps resolving the issue of processing stream resource that can not be seekable. 

**Of note, if your stream is not seekable you can still not detect its input BOM**


nyamsprod:
@clcastro87 I've improved the method names and align the UX to match #354  method names and signatures.

PR #354, opened 2019-08-28
Adding the ability to control skipping empty records by the Reader class

nyamsprod:
This PR adds the ability in the Reader class to control if empty records should be skipped or not. By default and to avoid BC break empty records are skipped. The result of `Reader::getRecords` will be controlled by the following methods.

- `Reader::preserveEmptyRecord`
- `Reader::skipEmptyRecord`
- `Reader::isEmptyRecordSkipped`


nyamsprod:
@clcastro87  this PR was prompted by your issue it adds the ability to also control if empty records should be skipped or not. If you could test it and tell me if that's also good for you. Again I should stress that those features will only be release end of september at least.

PR #356, opened 2019-09-11
mysql converter added

turkeryildirim:
This change makes possible to generate ready to run SQL strings from a well formatted and UTF8 encoded CSV file.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "php" : ">=7.0.10",
         "ext-dom" : "*",
         "ext-json" : "*",
-        "ext-mbstring" : "*"
+        "ext-mbstring" : "*",
+        "ext-iconv": "*"
     },
     "require-dev": {
         "ext-curl" : "*",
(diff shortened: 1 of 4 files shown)


nyamsprod:
@turkeryildirim thanks for the PR submission I would have liked to have a more in depth explanation as to why this is needed so that I would have had the chance to explain why I have some strong feeling against this addition.

First of all on the surface this is a great PR and I welcome these addition to the package however. Converting a CSV to MySQL statement should be made carefully when looking at first glance at your code I see that you made many assumptions about the SQL engine and about how this conversion should be done. Also I see some possible SQL injection in the code. 

That's the reason why I favor convertion to SQL being done by the developer who needs it on a case by case basis in order to mitigate these kind of issue and also to enable better maintenance of the current codebase.

I would suggest however to continue with your idea which should be done IMHO on an extension package that I would gladly promote once you've done creating it.

What do you think ?

Issue #360, opened 2019-11-04
Throw exceptions with context


dmlogic:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | yes

#### Proposal

Whenever an exception is thrown it is always a native `Exception`. This can make it hard to figure out what has gone wrong if the CSV handling is just part of a larger process.

I would like to see native `League\Csv\Exception` errors being throw to give some context, for example `StreamException` or `HeaderException`.

I'm very happy to do this work if there's a good chance of it being merged.

nyamsprod:
@dmlogic all exception thrown extends the League\Csv\Exception unless it is not possible for instance for the XMLConverter class. Maybe you could add more specific exceptions but the messages are already self explanatory unless I'm missing something. Either way I fail to see the BC breaks unless the newly added exceptions do not extend the marker exception or you want to throw on new context that currently don't throw

dmlogic:
> the messages are already self explanatory unless I'm missing something

They are, but because only one type of exception is thrown it is necessary to break up code into multiple stages of try/catch to infer the problem from context (I know there's a strong argument that I _should_ be doing that) or do a string comparison on the message.

I'd love to see more specific exceptions, which would probably need to be filed into `League\Csv\Exceptions`, hence the break. Although I guess `League\Csv\Exception` could stay where it is.

nyamsprod:
> I'd love to see more specific exceptions, which would probably need to be filed into `League\Csv\Exceptions`, hence the break. Although I guess `League\Csv\Exception` could stay where it is.

That's an enhancement not a BC break since the new Exceptions would be extending the current exception marker class ... 🤔 . Also there are already specific exception in the root directory like `CannotInsertRecord` exception following this pattern so a subnamespace is not really needed IMHO 😉

dmlogic:
OK - should I do a PR for a `HeaderRowException` or is this a "no"?

nyamsprod:
When will this exception be thrown? I'd rather have a `CannotReadRecord` exception and perhaps a `SyntaxError` but I'm not confident a specific exception for header is necessary

dmlogic:
I propose a straight swap-out for [this one](https://github.com/thephpleague/csv/blob/master/src/Reader.php#L323)

In my case I load a CSV from a stream and specify the header as row zero.

[...]

PR #361, opened 2019-11-07
Throw exceptions with context

dmlogic:
As per discussion in #360, this introduces additional exceptions to reduce the need to examine the error message string

### DIFF
--- a/src/AbstractCsv.php
+++ b/src/AbstractCsv.php
@@ -486,7 +486,7 @@ abstract class AbstractCsv implements ByteSequence
     public function addStreamFilter(string $filtername, $params = null): self
     {
         if (!$this->document instanceof Stream) {
-            throw new CannotAddStreamFilter('The stream filter API can not be used');
+            throw new StreamFilterSupportMissing('The stream filter API can not be used');
         }
 
         $this->document->appendFilter($filtername, $this->stream_filter_mode, $params);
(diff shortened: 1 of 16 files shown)


nyamsprod:
you should run `composer test` prior to submitting your commit. 

if you had some issue with coding style you can type 

```
./vendor/bin/php-cs-fixer fix -vvv --allow-risky=yes
```

src/Stream.php
```diff
@@ -243,7 +243,7 @@ public function appendFilter(string $filtername, int $read_write, $params = null
             return;
         }
 
-        throw new Exception(sprintf('unable to locate filter `%s`', $filtername));
+        throw new CannotAddStreamFilter(sprintf('unable to locate filter `%s`', $filtername));
```
nyamsprod: This should be a `InvalidArgument` exception. Stream Filtering is supported *but* the filter name is invalid.

src/AbstractCsv.php
```diff
@@ -486,7 +486,7 @@ public function setOutputBOM(string $str): self
     public function addStreamFilter(string $filtername, $params = null): self
     {
         if (!$this->document instanceof Stream) {
-            throw new Exception('The stream filter API can not be used');
+            throw new CannotAddStreamFilter('The stream filter API can not be used');
```
nyamsprod: I'm wondering if we should not name this one `MissingStreamFilterSupport` or `StreamFilterSupportMissing`

dmlogic: `StreamFilterSupportMissing` sounds good to me. Want me to change that?

nyamsprod: yes please

Issue #363, opened 2019-12-03
bom_match does not match UTF-32 LE


jmwebservices:
### Bug Report

| Information | Description |
|--------------|---------|
| Version | current  |
| PHP version | 7.3.12 |
| OS Platform |  Win |

#### Summary

`bom_match()` incorrectly returns `BOM_UTF16_LE` for strings with `BOM_UTF32_LE`. This is due to the order in which these constants are defined in `League\Csv\ByteSequence`. 

#### Standalone code, or other way to reproduce the problem

```
var_dump( \League\Csv\bom_match( \League\Csv\ByteSequence::BOM_UTF32_LE ) === \League\Csv\ByteSequence::BOM_UTF32_LE );
```

#### Expected result

bool(true)

#### Actual result

bool(false)

PR #364, opened 2019-12-03
Reorder ByteSequence constants

jmwebservices:
Fix for #363

### DIFF
--- a/src/ByteSequence.php
+++ b/src/ByteSequence.php
@@ -17,6 +17,16 @@ namespace League\Csv;
 interface ByteSequence
 {
     /**
+     * UTF-32 BE BOM sequence.
+     */
+    const BOM_UTF32_BE = "\x00\x00\xFE\xFF";
+
+    /**
+     * UTF-32 LE BOM sequence.
+     */
+    const BOM_UTF32_LE = "\xFF\xFE\x00\x00";
+
+    /**
      *  UTF-8 BOM sequence.
      */
     const BOM_UTF8 = "\xEF\xBB\xBF";
@@ -30,14 +40,4 @@ interface ByteSequence
      * UTF-16 LE BOM sequence.
      */
     const BOM_UTF16_LE = "\xFF\xFE";
-
-    /**
-     * UTF-32 BE BOM sequence.
-     */
-    const BOM_UTF32_BE = "\x00\x00\xFE\xFF";
-
-    /**
-     * UTF-32 LE BOM sequence.
-     */
-    const BOM_UTF32_LE = "\xFF\xFE\x00\x00";
 }
--- a/tests/ByteSequenceTest.php
+++ b/tests/ByteSequenceTest.php
@@ -43,15 +43,15 @@ class ByteSequenceTest extends TestCase
                 'sequence' => 'foo bar',
                 'expected' => '',
             ],
-            'UTF8 BOM sequence' => [
+            'UTF-8 BOM sequence' => [
                 'sequence' => chr(239).chr(187).chr(191),
                 'expected' => ByteSequence::BOM_UTF8,
             ],
-            'UTF8 BOM sequence at the start of a text' => [
+            'UTF-8 BOM sequence at the start of a text' => [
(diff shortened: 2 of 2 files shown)


src/ByteSequence.php
```diff
@@ -17,27 +17,27 @@
 interface ByteSequence
 {
     /**
-     *  UTF-8 BOM sequence.
+     * UTF-32 BE BOM sequence.
```
nyamsprod: the re-ordering is not the issue, the issue is in the implementation of `bom_match` the function should be updated by sorting the constant using their value length.

nyamsprod:
@jmwebservices thanks for the PR. This is a good start but to resolve the issue, the `bom_match` function should be instead changed to something along these lines 

```php

<?php

    function bom_match(string $str): string
    {
        static $list;
        if (null === $list) {
            $list = (new ReflectionClass(ByteSequence::class))->getConstants();

            rsort($list, SORT_NATURAL | SORT_FLAG_CASE);
        }

        foreach ($list as $sequence) {
            if (0 === strpos($str, $sequence)) {
                return $sequence;
            }
        }
        
        return '';
    }

```

I have no tested the result but this is the way to go, Changing the order in the interface makes it part of a specific implementation and that defeat the purpose of an interface.

src/ByteSequence.php
jmwebservices: I too thought relying on a definition order in the interface was flaky but it was the path of least resistance. And, the test would guard it from becoming out of order in the future.

jmwebservices:
Would you be interested in the following?
1. Removing the Reflection magic and unrolling the tests so that each BOM sequence is tested individually? I believe it would reduce the LOC and make it clearer.
2. Operating on the first 4 bytes of `$str` since `strpos` will continue checking for the 1st occurrence through the end of `$str`.

nyamsprod:
@jmwebservices 

1/ the reflection magic is done for maintenance and speed:

- if we need to add more constants the `bom_match` code does not need any update. 
-  a static local variable is used = the calculation is done only once, the first time the function is called, on subsequent calls nothing it done or executed since the value are already present and hopefully after your PR sorted 😉 

_So I would be against changing this part of the code._

2/ some tests should be conducted or some literature (with a link) should be provided to see if  `substr($long_str, 0, 10)` does require or not the string full length to operate.

If the full length is still computed then I see no gain in doing so 
If not yes please go ahead.. It's an optimization that can be handy for large CSV file IMHO

jmwebservices:
> 2/ some tests should be conducted or some literature (with a link) should be provided to see if `substr($long_str, 0, 10)` does require or not the string full length to operate.
> 
> If the full length is still computed then I see no gain in doing so
> If not yes please go ahead.. It's an optimization that can be handy for large CSV file IMHO

I am getting crazy performance improvements. On my machine, using `substr()` to `strpos()` the first 4 bytes takes about 1% of the time than if the entire string is checked. Please verify for yourself.

https://gist.github.com/jmwebservices/3cd9cb273ec52f835a6dd13d62b2bc76

jmwebservices:
This optimization may be unccessary since a quick search over the code base indicates that `bom_match()`  is only called once where the passed string is at most 4 bytes.

https://github.com/thephpleague/csv/blob/42c8916bd02d05169ea31a9479cc789a082cd93f/src/AbstractCsv.php#L228

nyamsprod:
@jmwebservices are you gonna modify the function or not else I can take over your PR and merge the changes so I can release before the end of the year 🤔🤔😉

jmwebservices:
> @jmwebservices are you gonna modify the function or not else I can take over your PR and merge the changes so I can release before the end of the year
I was waiting on your reply on my last comment about whether the optimization is needed at all. But yes, please go ahead and modify to address the bug. The end of year has me swamped.

Thanks for your contribution!

nyamsprod:
@jmwebservices sorry I thought by your last answer that you were not going to add the optimization since its useless in the CSV codebase 😢 . I'll still mention your contribution in the changelog don't worry about that 😉

nyamsprod:
this merge is closed and I'll be using #365 instead

PR #369, opened 2019-12-19
Corrected use of undefined variable in docs

unguul:
The `$record` variable doesn't exist but it is referenced in the example so readers could potentially get confused. I also removed an extra "s" from the nearby `includeEmptyRecords()` call.


nyamsprod:
@unguul thanks for improving the documentation 👍

PR #371, opened 2020-01-01
Added option to read zip compressed CSV file

giriannamalai:
When a zipped CSV file found, which is ends with ".zip" extension.
We extract it to temp folder and send that path instead of real.

### DIFF
--- a/src/Reader.php
+++ b/src/Reader.php
@@ -86,6 +86,28 @@ class Reader extends AbstractCsv implements Countable, IteratorAggregate, JsonSe
      */
     public static function createFromPath(string $path, string $open_mode = 'r', $context = null)
     {
+        # Endswith zip extension for compressed file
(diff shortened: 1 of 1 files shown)


nyamsprod:
@giriannamalai  thanks for the PR submission I would have liked to have a more in depth explanation as to why this is needed so that I would have had the chance to explain **why it is not**.

First of all on the surface this is a great PR and I welcome these addition to the package, however, this feature is already supported by the current codebase through stream filter usage. Therefore this addition is not needed at all. 

If you still want to use the ZipArchive class and it's inner dependency on the zip extension then I would suggest creating a small package around the League\Csv codebase to compress and decompress the CSV file prior to or after it's usage by the current codebase.

Again, thanks for the contribution and I'll still review any other submission against this library from you.

PR #396, opened 2020-08-31
CSV delimiter autodetect method

fazzinipierluigi:


nyamsprod:
@fazzinipierluigi thanks for the PR. 

I would like to understand the added value of adding this method versus using the documented function ? 

https://csv.thephpleague.com/9.0/connections/controls/#detecting-the-delimiter-character

fazzinipierluigi:
in my opinion the method is more consistent than it is in the class and does not return a statistical analysis of all the inserted delimiters but if one is found this is set, making it easier and more immediate to use.
It's just a practical motivation. =)

nyamsprod:
Understood but your implementation may lead to false positive and incorrect results. This is why this function which once was in the public API of the Reader class was removed from it. To clearly state that the response should be taken with a grain of salt.

nyamsprod:
@fazzinipierluigi   I will reject your PR on the ground that this is an old public API method previously called `detectDelimiterList` that was deprecated because of the false sense of certitude that it gives to users.

PR #406, opened 2020-12-10
do not allow for output during tests

IonBazan:
This PR restricts output during test and suppressed existing tests with output.
It also uses `$this->expectException()` instead of `self::expectException()` calls as the method is not static.


nyamsprod:
@IonBazan can I add this PR to the next release with support for PHP8 ?

IonBazan:
@nyamsprod yes, it’s just tests improvement 😉

PR #415, opened 2021-03-15
Improve PHPUnit assertion and fixtures

peter279k:
# Changed log

- Using the `assertContains` to assert expected contains the specific  key.
- According to the [official PHPUnit doc](https://phpunit.readthedocs.io/en/9.5/fixtures.html#more-setup-than-teardown), it should be `protected function setUp` and `protected function tearDown` methods.


nyamsprod:
@peter279k your fix is not passing the coding style check in PHP8.0. You need to remove thhe useless `use in_array` declaration.
After which I'll merge the PR

peter279k:
Fixed :).

PR #421, opened 2021-04-16
fix  writer filter

cpj555:
self -> static


nyamsprod:
@cpj555 thanks for the fix

nyamsprod:
9.7.1 is released with your fix

PR #428, opened 2021-06-16
Enhancement carying about array index writter

LeoDevSecOps:
---
name: 🎉 Insertion care about key index
about: Here is an implementation that supports inserting rows with an auto-generated field. In case you want to add a row and the array that you pass does not have indexes that follow each other strictly, it happens if a column of the row has no value, we do not put a key/value to this index.  
---

  * submitting a new feature: target the master branch.

## New Feature

_(Fill in the relevant information below to help triage your issue.)_

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

### Proposal

_(Please explain in plain english the feature you would like to see implemented.)_

## Checks before submitting

* [x] Be sure that there isn't already a pull request about this. See: [Pull requests](https://github.com/thephpleague/csv/pulls)
* [x] This pull request is about 1 new feature and nothing more.
* [x] Be sure to always target the lowest supported PHP version.
* [x] Be sure to include tests.
* [x] Be sure to have run `composer test` before submitting the fix.
* [x] Be sure to update the `CHANGELOG.md`
* [x] Be sure to update the documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,1161 +6,1159 @@ All Notable changes to `Csv` will be documented in this file
 
 ### Added
 
-- None
+-   Insertion ro carying about key index to refer csv field.
 
 ### Deprecated
 
-- None
+-   None
 
 ### Fixed
 
-- fix writer filter [#421](https://github.com/thephpleague/csv/pull/421) by [LosingBattle](https://github.com/cpj555)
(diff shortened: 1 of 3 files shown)


nyamsprod:
@LeoDevSecOps thanks for the proposed addition. Can you tell me the added value of this new feature versus relying on the currently existing [Writer::addFormatter](https://csv.thephpleague.com/9.0/writer/#writeraddformatter) feature ?

PR #430, opened 2021-06-29
Fix old link for php-cs-fixer landing page

andrew-demb:


nyamsprod:
Thanks for the update link 👍🏾

PR #436, opened 2021-09-30
Don't exclude docs from packages

georgringer:
Having the documentation locally after a composer install is very useful.


nyamsprod:
@georgringer thank for using the library. But I think that the documentation should either be cloned from GitHub and is already accessible even without download on the documentation website. Adding it to the build seems strange to me if you can use composer to get the package you should be able to get to the documentation website no ?

georgringer:
I just like having all in my IDE but I fully understand your point. Feel free to reject issue and pr

PR #452, opened 2021-11-29
Use single quote to escape formulas

robertfausk:
... and add tab ```\t``` and carriage return ```\r``` to escaped values.

As already used in symfony/serializer (https://symfony.com/blog/cve-2021-41270-prevent-csv-injection-via-formulas)
and as OWASP suggests in https://owasp.org/www-community/attacks/CSV_Injection

Credits for finding this issue in thephpleague/csv goes to @Lehmhub :dancing_men:


nyamsprod:
@robertfausk thanks for the PR I will release a new version ASAP

PR #456, opened 2022-01-15
Minor fixes

george-zakharov:
Hi!

I've made some minor style fixes.
Regards

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -1,15 +1,15 @@
 <?php
 
-require __DIR__.'/src/functions_include.php';
+require __DIR__ . '/src/functions_include.php';
 
-spl_autoload_register(function ($class) {
+spl_autoload_register(static function ($class) {
 
     $prefix = 'League\Csv\\';
     if (0 !== strpos($class, $prefix)) {
         return;
     }
 
-    $file = __DIR__.'/src/'.str_replace('\\', '/', substr($class, strlen($prefix))).'.php';
+    $file = __DIR__ . '/src/' . str_replace('\\', '/', substr($class, strlen($prefix))) . '.php';
     if (!is_readable($file)) {
         return;
     }
--- a/autoload.php
+++ b/autoload.php
@@ -1,15 +1,15 @@
 <?php
 
-require __DIR__ . '/src/functions_include.php';
+require __DIR__.'/src/functions_include.php';
 
-spl_autoload_register(static function ($class) {
+spl_autoload_register(function ($class) {
 
     $prefix = 'League\Csv\\';
     if (0 !== strpos($class, $prefix)) {
         return;
     }
 
-    $file = __DIR__ . '/src/' . str_replace('\\', '/', substr($class, strlen($prefix))) . '.php';
+    $file = __DIR__.'/src/'.str_replace('\\', '/', substr($class, strlen($prefix))).'.php';
     if (!is_readable($file)) {
         return;
     }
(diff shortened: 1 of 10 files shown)


autoload.php
```diff
@@ -1,15 +1,15 @@
 <?php
 
-require __DIR__.'/src/functions_include.php';
+require __DIR__ . '/src/functions_include.php';
 
-spl_autoload_register(function ($class) {
+spl_autoload_register(static function ($class) {
```
nyamsprod, line 5: the addition of `static` is more of a micro-optimization than anything else I would not add it here

src/AbstractCsv.php
```diff
@@ -323,7 +323,7 @@ public function output(string $filename = null): int
 
         echo $this->output_bom;
 
-        return strlen($this->output_bom) + (int) $this->document->fpassthru();
+        return strlen($this->output_bom) + (int)$this->document->fpassthru();
```
nyamsprod, line 326: Why removing the space ??

src/CharsetConverter.php
```diff
@@ -94,7 +94,7 @@ protected static function filterEncoding(string $encoding): string
             return $encoding_list[$key];
         }
 
-        throw new OutOfRangeException('The submitted charset '.$encoding.' is not supported by the mbstring extension.');
+        throw new OutOfRangeException('The submitted charset ' . $encoding . ' is not supported by the mbstring extension.');
```
nyamsprod, line 97: Why adding spaces ? Did you run `composer phpcs` ?

src/TabularDataReader.php
```diff
@@ -48,7 +48,7 @@ public function getIterator(): Iterator;
     /**
      * Returns the header associated with the tabular data.
      *
-     * The header must contains unique string or is an empty array
+     * The header must contain unique string or to be an empty array
```
nyamsprod, line 51: good catch

nyamsprod:
@george-zakharov I like the typo fixes but I am a bit against the coding style changes you added in the PR. Those coding style exists since v9 and even if I want to change some I will probably update them for v10 whenever that one comes alive but not before to stay as consistent as possible.
If you could update your PR and only submit the typo changes I would gladly merge them ASAP

autoload.php
george-zakharov: Okay, I see.

src/AbstractCsv.php
george-zakharov: Well, it's an interesting thing :)
I did it according to my habbits. I've checked PSR once again and haven't find the any clarity about such situation.
I think that [here](https://github.com/php-fig-rectified/fig-rectified-standards/issues/13#issuecomment-324653158) is a pretty good explanation why. I agree with it.
Anyway, according to the inner code style of this project my fix is not neccessary.

src/CharsetConverter.php
george-zakharov: Well, I've checked standards and haven't find there any consistency about spaces around concatenation dot.
My point is readability matters.
Anyway, I think we don't need my fix here :)

george-zakharov:
> 

Okay, I'll fix it :)

george-zakharov:
@nyamsprod did it :)

nyamsprod:
@george-zakharov thanks. By the way now that I saw your answer I get it. League CSV uses PSR-2 as the basis of it's coding style + some additions by me. It was done prior to PSR-12 or PSR-2R was even created This explain more my why's 😉 . Anyway I will review the package coding style when the next major version is released. Thanks for the PR in any case.

george-zakharov:
@nyamsprod thank you too!
I will try to contribute further :)

PR #474, opened 2022-09-15
replace `foreach` with `implode()`

SubhanSh:


src/AbstractCsv.php
```diff
@@ -290,10 +290,7 @@ public function getContent(): string
      */
     public function toString(): string
     {
-        $raw = '';
-        foreach ($this->chunk(8192) as $chunk) {
-            $raw .= $chunk;
-        }
+        $raw = implode('', $this->chunk(8192));
```
nyamsprod, line 293: the method `chunck` returns a `Generator` I am not aware `implode` supports them.

SubhanSh: It seems it doesn't support :(

PR #478, opened 2022-10-21
Add sensible open mode default for Writer

weshooper:
The documentation gives (at https://csv.thephpleague.com/9.0/writer/#inserting-records) the following example:

```php
$writer = Writer::createFromPath('/path/to/saved/file.csv', 'w+');
```

The second parameter, 'w+' is needed because the default open mode in `AbstractCsv.php` is 'r+', see https://github.com/thephpleague/csv/blob/master/src/AbstractCsv.php#L109

Since the purpose of the `Writer` class is to write, it seems a more sensible open mode default would be write (i.e. 'w+') rather than read.

This allows us to simplify usage to:

```php
$writer = Writer::createFromPath('/path/to/saved/file.csv');
```

Additionally, the PHP manual suggests using 'b' (at https://www.php.net/manual/en/function.fopen.php) so went for 'w+b' for a default.

### DIFF
--- a/src/Writer.php
+++ b/src/Writer.php
@@ -32,6 +32,14 @@ class Writer extends AbstractCsv implements TabularDataWriter
     protected ?int $flush_threshold = null;
 
     /**
+     * {@inheritdoc}
+     */
+    public static function createFromPath(string $path, string $open_mode = 'w+b', $context = null)
+    {
+        return parent::createFromPath($path, $open_mode, $context);
+    }
+
+    /**
      * Returns the current newline sequence characters.
      */
     public function getNewline(): string


nyamsprod:
@weshooper thanks for the PR can you elaborate on the proposed changed and why it should be used as a default ?

weshooper:
@nyamsprod thanks, I've updated the PR description.

nyamsprod:
@weshooper with the context given I understand where you are coming from and this issue has already been tackled in the past. In summary you are introducing a small yet sensible BC break. so this "fix" was differ for the next major release whenever that one will come.

Last but not least while I fully agree that the default open mode for the Writer should be changed .. I am a bit skeptical on this addition of the `b` mode to it. The reason being that a CSV file is not binary. But again you might prove me wrong.

Hope this answer your question

nyamsprod:
This PR will be closed and the issue will be tackle when version 10 development starts.

weshooper:
Thanks @nyamsprod

Issue #508, opened 2023-10-30
Add object mapping


nyamsprod:
## Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

## Introduction

One of the most requested feature is to add the ability to map CSV record to object.
Currently the package only allows:

- `array<array-key, string|null>` 
-  typed array with broader support for record value type via the use of `Reader::addFormatter`.

## Proposal

We need a mechanism the allow converting array to object. The mechanism should be:

- transparent
- ease to use and maintain
- extensible to allow for user defined rules on casting

In order to do so we propose to introduce a new `map` method to `TabularDataReader` with the following signature:

```php
TabularDataReader::map(string $class): Iterator;
```

Let's assume the following CSV document.

```php
$weather = <<<CSV
date,temperature,place
2011-01-01,1,Galway
2011-01-02,-1,Galway
2011-01-03,0,Galway
2011-01-01,6,Berkeley
2011-01-02,8,Berkeley
2011-01-03,5,Berkeley
CSV;
```

And the following DTO and Enum classes.

```php
class Weather
{
    private DateTime $createdAt;

    public function __construct(
        #[League\Csv\Attribute\Column(offset:'temperature')]
        public readonly int $temperature,
        #[League\Csv\Attribute\Column(offset:2, cast: CastToEnum::class)]
        public readonly Place $place,
    ) {
    }

    public function getDate(): DateTime
    {
        return $this->createdAt;
    }

    #[League\Csv\Attribute\Column(
        offset: 'date',
        cast: CastToDate::class,
        castArguments: ['format' => '!Y-m-d', 'timezone' => 'Africa/Kinshasa']
    )]
    public function setDate(DateTime $createdAt): void
    {
        $this->createdAt = $createdAt;
    }
}

enum Place: string
{
    case Galway = 'Galway';
    case Berkeley = 'Berkeley';
}
```

Using the new `map` method we should be able to do the following

```php
$csv = Reader::createFromString($weather);
$csv->setHeaderOffset(0);
$record = [...$csv][0];
$weather = [...$csv->map(Weather::class)][0];

var_dump($record);
// returns ['date' => '2011-01-01', 'temperature' => '1', 'place' => 'Galway']

echo $weather->getDate()->format('d/m/Y'); // display '01/01/2011'
echo $weather->place->value; // display the enum value 'Galway';
echo $weather->temperature;  // display 1 as an integer
```

[...]

PR #509, opened 2023-10-30
#508 Adding Object Mapping

nyamsprod:
Implementation object mapping as described in #508


docs/9.0/reader/tabular-data-reader.md
```diff
@@ -117,6 +117,203 @@ and its value will represent its header value.
 This means that you can re-arrange the column order as well as removing or adding column to the
 returned iterator. Added column will only contain the `null` value.
 
+### Mapping recrods to objects
```
hskrasek, line 120: Small typo here, records to recrods

PR #525, opened 2024-05-30
Docs: Fix sample code in documentation to use `getRecords()` instead of `getHeader()`

meihei3:
The validation for duplicated headers requires calling `Reader::computeHeader()`, but `Reader::getHeader()` does not invoke this method. As a result, the current sample code does not produce the expected `SyntaxError`.

I have updated the sample code to use `getRecords()`, which correctly triggers the `SyntaxError` for duplicate headers.

Thanks.

### DIFF
--- a/docs/9.0/reader/index.md
+++ b/docs/9.0/reader/index.md
@@ -225,7 +225,7 @@ $csv = Reader::createFromPath('/path/to/file.csv', 'r');
 $csv->fetchOne(0); //returns ['field1', 'field2', 'field1', 'field4']
 $csv->setHeaderOffset(0); //valid offset but the record contain duplicates
 $header_offset = $csv->getHeaderOffset(); //returns 0
-$header = $csv->getHeader(); //throws a SyntaxError exception
+$records = $csv->getRecords(); //throws a SyntaxError exception
 ```
 
 <p class="message-info">Starting with <code>9.7.0</code> the <code>SyntaxError</code> exception thrown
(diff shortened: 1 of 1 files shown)


nyamsprod:
@meihei3 The code you are trying to update is related to the header and is correct AFAIK. `getHeader` will throw an exception if the header offset value set using `setHeaderOffset` is invalid because the package does a lazy evaluation of the header offset, the offset is really calculated once `getHeader` is called or when `getRecords` is used.

meihei3:
@nyamsprod  Thank you for your feedback. I understand that `getHeader` will throw an exception if the header offset value set using `setHeaderOffset` is invalid due to lazy evaluation. However, the issue I encountered is specific to duplicated header entries.

While `getHeader` does call `setHeaderOffset`, `setHeaderOffset` does not call `SyntaxError::dueToDuplicateHeaderColumnNames($header)`. This can be seen in the relevant code here: [Reader.php#L76-L90](https://github.com/thephpleague/csv/blob/e201a39ea45da21d84ce4103a42723753ca5b52c/src/Reader.php#L76-L90) and [Reader.php#L562](https://github.com/thephpleague/csv/blob/e201a39ea45da21d84ce4103a42723753ca5b52c/src/Reader.php#L562).

As a result, the current sample code does not produce the expected `SyntaxError` for duplicated headers. In contrast, `getRecords()` invokes the necessary method to detect duplicate columns and correctly trigger the `SyntaxError`.

Thanks.

nyamsprod:
@meihei3 the documentation is right but the implementation has a bug!! So I will convert your ticket into a bug issue that I will fix ASAP!! So the code needs to be changed not the documentation

nyamsprod:
Now that I am thinking a bit about it you may be right. When you choose a row to be the header, if that row exists it should be returned as is. You did select the row afterall. What is triggering the error is **when** you try to use that row as a header. So should the exception be thrown when you still have not decided to use the header or when you are in fact using the error. If it's the former then it is a bug in the code ... if its the latter then it is a bug in the documentation 🤷🏾

meihei3:
@nyamsprod  Thank you! However, I don't have the permissions to merge this PR, so could you please merge it on my behalf?

PR #544, opened 2024-12-17
Remove stream filters in Stream destructor only if they are still valid resources

crocodele:
When a `Reader` or `Writer` has a stream filter and the handle to the stream is closed manually in code, there's a fatal error in the `Stream` destructor.

To reproduce: in current `master` branch, the test case in the PR fails with a fatal error in the `Stream` destructor:

```
PHP Fatal error:  Uncaught TypeError: stream_filter_remove(): supplied resource is not a valid stream filter resource in /path/to/example/vendor/league/csv/src/Stream.php:84
Stack trace:
#0 /path/to/example/vendor/league/csv/src/Stream.php(84): stream_filter_remove()
#1 [internal function]: League\Csv\Stream->League\Csv\{closure}()
#2 /path/to/example/vendor/league/csv/src/Stream.php(84): array_walk_recursive()
#3 [internal function]: League\Csv\Stream->__destruct()
#4 {main}
  thrown in /path/to/example/vendor/league/csv/src/Stream.php on line 84
```

### DIFF
--- a/src/ReaderTest.php
+++ b/src/ReaderTest.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace League\Csv;
 
 use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\DoesNotPerformAssertions;
 use PHPUnit\Framework\Attributes\Group;
 use SplFileObject;
 use SplTempFileObject;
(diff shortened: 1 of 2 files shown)


nyamsprod:
Thanks for finding and fixing the bug. Nice catch. I will merge the PR but I update or improve if needed your fix.

PR #548, opened 2025-01-09
Fix Comparison::CONTAINS to check is_string before str_compare

cage-is:
In Comparison there is a line that says 

```php
self::Contains => str_contains($subject, $reference), /* @phpstan-ignore-line */
```

Since this is checking `$subject` with `str_contains` it should check first if it is a string. This seems to have been handled in other various comparisons already (e.g. `self::Regexp => is_string($subject) && 1 === preg_match($reference, $subject), /* @phpstan-ignore-line */`)

```
TypeError: str_contains(): Argument #1 ($haystack) must be of type string, null given
/var/www/src/Query/Constraint/Comparison.php:102
/var/www/src/Query/Constraint/Column.php:78
/var/www/src/Query/Constraint/Criteria.php:47
/var/www/src/Query/Constraint/Criteria.php:108
/var/www/src/ResultSet.php:395
/var/www/src/StatementTest.php:260
```


nyamsprod:
thanks for the bugfix and the nice catch! I will merge your contribution 👍🏾

nyamsprod:
I have updated your test case the assertSame was not mandatory an assertCount is enough as the `ResultSet` implements the Countable interface which means that iteration is done which highlights your fix 😉

Issue #560, opened 2025-03-21
Add field name to TypeCastingFailed exception


OliverBeckwith:
## Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no

### Proposal

The current `TypeCastingFailed` exception doesn't provide details about which field is failing. It would be nice if the exception message included the column name (pulled from the `MapCell` attribute, or falling back to just the field name in the class). When you have a CSV with a lot of columns, it can be difficult to determine which value is currently failing the type casting, and especially so when this error is exposed to non-technical users that can't look at the code.

I have had a go at providing the information as metadata in the exception object here: https://github.com/thephpleague/csv/compare/master...OliverBeckwith:thephpleague_csv:type_cast_exception_metadata, but I don't particularly like the solution

## Checks before submitting

* [X] Be sure that there isn't already an issue about this. See: [Issues list](https://github.com/thephpleague/csv/issues)
* [X] Be sure that there isn't already a pull request about this. See: [Pull requests](https://github.com/thephpleague/csv/pulls)
* [X] Tell if you have the option to provide the code for this proposal.
* [X] This issue is about 1 feature proposal and nothing more.
* [X] The issue has a descriptive title. For example: "Add JSON render option".

nyamsprod:
Adding the field name is a bit tricky while adding the object property or argument name is more straightforward. Can you test the PR to see if the output is better. 🤔

OliverBeckwith:
Hi. That is definitely an improvement. In the PR, since you're getting the name from the reflection property, would it not therefore be possible to do the same logic of getting the MapCell attribute if present and using that as the field name for the exception? 
Something like (haven't actually tested this):
```php
$attributes = $reflectionProperty->getAttributes(MapCell::class, ReflectionAttribute::IS_INSTANCEOF);
if (count($attributes) && is_string($column = $attributes[0]->newInstance()->column)) {
    $variableName = $column;
} else {
    $variableName = $reflectionProperty->getName();
}
```

Since I see you doing it in a lot of different classes, perhaps a helper that takes in the reflection property and returns the field name? I am just throwing ideas out, this is already a massive improvement thank you

nyamsprod:
PR updated taking into account your remarks

nyamsprod:
PR merge into the main branch

PR #561, opened 2025-03-21
Improve TypeCastingFailed message

nyamsprod:
Resolve issue #560 

We can not add the record field *BUT* we can add the object property so this should give enough information to help debugging the error.

### DIFF
--- a/src/Serializer/TypeCastingFailed.php
+++ b/src/Serializer/TypeCastingFailed.php
@@ -16,39 +16,54 @@ namespace League\Csv\Serializer;
 use RuntimeException;
 use Throwable;
 
+use function is_int;
+
 final class TypeCastingFailed extends RuntimeException implements SerializationFailed
 {
-    public static function dueToNotNullableType(string $type, ?Throwable $exception = null, ?string $variableName = null): self
-    {
-        $message = match ($variableName) {
-            null => 'The `null` value can not be cast to an `'.$type.'`; the property type is not nullable.',
-            default => 'The `null` value can not be cast to an `'.$type.'` for the object property `'.$variableName.'``; the property type is not nullable.',
-        };
+    public readonly ?TypeCastInfo $info;
 
-        return new self($message, 0, $exception);
+    public function __construct(string $message, int $code = 0, ?Throwable $previous = null, ?TypeCastInfo $info = null)
+    {
+        parent::__construct(self::format($message, $info), $code, $previous);
+
+        $this->info = $info;
     }
 
-    public static function dueToInvalidValue(mixed $value, string $type, ?Throwable $previous = null, ?string $variableName = null): self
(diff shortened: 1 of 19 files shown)


src/Serializer/TypeCastingFailed.php
```diff
@@ -18,22 +18,37 @@
 
 final class TypeCastingFailed extends RuntimeException implements SerializationFailed
 {
-    public static function dueToNotNullableType(string $type, ?Throwable $exception = null): self
+    public static function dueToNotNullableType(string $type, ?Throwable $exception = null, ?string $variableName = null): self
     {
-        return new self('The `null` value can not be cast to an `'.$type.'`; the property type is not nullable.', 0, $exception);
+        $message = match ($variableName) {
+            null => 'The `null` value can not be cast to an `'.$type.'`; the property type is not nullable.',
+            default => 'The `null` value can not be cast to an `'.$type.'` for the object property `'.$variableName.'``; the property type is not nullable.',
```
OliverBeckwith, line 25: Spotted an extra ` here after the $variableName.

Unrelated: is there any chance you could take out the "an" before the $type, feels grammatically a bit odd to me.

src/Serializer/TypeCastingFailed.php
```diff
@@ -18,22 +18,37 @@
...
+        return new self($message, 0, $previous);
     }
 
-    public static function dueToUndefinedValue(string|int $offset): self
+    public static function dueToUndefinedValue(string|int $offset, ?string $variableName = null): self
     {
-        return new self('Unable to cast the record value; Missing value was for  offset`'.$offset.'.', 0);
+        $message = match ($variableName) {
+            null => 'Unable to cast the record value; Missing value was for offset`'.$offset.'.',
+            default => 'Unable to cast the record value; Missing value was for offset`'.$offset.' and for the object property `'.$variableName.'`.',
```
OliverBeckwith, line 49: Unrelated: Should there be a closing backtick around the $offset?

nyamsprod:
I updated the PR it is a bit more complex but the message should now be even clearer.

If the error comes from a setter method you will get something like this 

> Fatal error: Uncaught League\Csv\Serializer\TypeCastingFailed: Casting the method `setReview()` first argument `foobar` using the record field `review` failed; The `null` value can not be cast to a `string`; the property type is not nullable.

if an offset is used instead you will get the following:

> Fatal error: Uncaught League\Csv\Serializer\TypeCastingFailed: Casting the property `tags` using the record field offset `5` failed; The `null` value can not be cast to a `string`; the property type is not nullable.

The current message is no longer changed but it is prefixed with a more detail information about the context around the type casting failure.

OliverBeckwith (approved):
Works great. These error messages are much more helpful now

PR #566, opened 2025-07-20
match denormalized properties, fixes #565

### LINKED ISSUE #565: allow denormalized matching in getRecordsAsObjects
## Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| BC Break    | no (if new option is added)

### Proposal

My csv has capitals and spaces in the header, I'd like to be able to map it to programmer-friendly variables without using MapCell..

Title","url","wam_image","Accession Number","Cataloguer","Collection","Object Description","Keywords","Location","Notes","Registration Number","Subject","Thumbnail","Condition","Creator","Digital ID","Medium","Medium Format","Notes2","Original Legal Broker","State"

### Example

```
class Obj 
{


    // matches 
        #[MapCell('ObjectDescription')]
        #[MapCell('Object Description')]
        #[MapCell('Object_Description')]
        #[MapCell('objectdescription')]
        #[MapCell('object_description')]

        public ?string $objectDescription=null;

```

The Symfony ProjectAccessor uses pretty loose rules for getting and setting, and I find it quite helpful.

On a related note, MapCell can't be repeated.  But I want to match any of those, so that my DTO can be resuable for the myriad of csv headers.  (If only our user could be consistent!)

Currently I'm using getRecords($mappedHeaders) and cleaning up the headers beforehand.  But the record mapper is so much more powerful that I'd like to use it.

tacman:
instead of an exact property match, use denormalized columns:

```

        $records = [
            [
                'observedOn' => '2023-10-30',
                'temperature' => '-1.5',
                'place' => 'Abidjan',
            ],
            [
                'observed_on' => '2023-10-30',
                'temperature' => '-3',
                'place' => 'Abidjan',
            ],
            [
                'observed On' => '2023-10-30',
                'temperature' => '-3',
                'place' => 'Abidjan',
            ],
            [
                'observedon' => '2023-10-30',
                'temperature' => '-3',
                'place' => 'Abidjan',
            ],
        ];

```

### DIFF
--- a/src/Serializer/Denormalizer.php
+++ b/src/Serializer/Denormalizer.php
@@ -386,8 +386,11 @@ final class Denormalizer
         return false; // No match
     }
 
-    private static function normalizeKey(string $key): string
+    private static function normalizeKey(?string $key): ?string
     {
+        if (is_null($key)) {
+            return null;
+        }
         // Remove spaces and underscores
         $key = str_replace([' ', '_'], '', $key);
 
--- a/src/Serializer/Denormalizer.php
+++ b/src/Serializer/Denormalizer.php
@@ -341,7 +341,8 @@ final class Denormalizer
             }
 
             /** @var int<0, max>|false $index */
-            $index = array_search($offset, $propertyNames, true);
+            $index = self::normalizedMatch($offset, $propertyNames);
+
             if (false === $index) {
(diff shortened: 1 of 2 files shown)


nyamsprod:
See #565 for reasons to close this PR

PR #573, opened 2026-02-04
No need to do second `instanceof` in `insertAll()` when first satisfied

weshooper:
I guess inconsequentially more performant, I've not checked. Just easier for brains to read.

### DIFF
--- a/src/Writer.php
+++ b/src/Writer.php
@@ -119,9 +119,7 @@ class Writer extends AbstractCsv implements TabularDataWriter
     {
         if ($records instanceof TabularDataProvider) {
             $records = $records->getTabularData();
-        }
-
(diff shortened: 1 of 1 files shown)


src/Writer.php
```diff
@@ -119,9 +119,7 @@ public function insertAll(TabularDataProvider|TabularData|iterable $records): in
     {
         if ($records instanceof TabularDataProvider) {
             $records = $records->getTabularData();
-        }
-
-        if ($records instanceof TabularData) {
+        } elseif ($records instanceof TabularData) {
```
nyamsprod: This is not correct because 
`$records->getTabularData()` returns an instance of `TabularData` so you are bypassing the second check. But if reading is the issue there is another way to do that I will think about it

Issue #579, opened 2026-06-22
Use SortDirection from PHP 8.6 internally in Column class.


okorneliuk:
See https://wiki.php.net/rfc/sort_direction_enum

How about using [SortDirection](https://wiki.php.net/rfc/sort_direction_enum) instead of ordering constants in [Column](https://github.com/thephpleague/csv/blob/master/src/Query/Ordering/Column.php)? We can [polyfill](https://github.com/symfony/polyfill-php86) it though. This approach is already adopted by Doctrine: https://github.com/doctrine/collections/pull/542

nyamsprod:
The feature is merged in the main branch

PR #580, opened 2026-06-25
Utilized SortDirection enum from PHP 8.6

okorneliuk:
Related to #579

Added a polyfill of upcoming PHP 8.6, used `SortDirection` enum from it instead of private constants in `Column`. Slightly extended public API so it is possible to use `SortDirection` directly in `sortOn` method.


nyamsprod:
@okorneliuk thanks for the PR. While it would be nice to have this inside the package. The CSV package is meant to be used without dependencies so the best course of action would be to update this PR by including the polyfill directly inside the package.

polyfill/sort_direction.php
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * (c) Ignace Nyamagana Butera <nyamsprod@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+if (!enum_exists('SortDirection', false)) {
```
nyamsprod: I would simply use `PHP_VERSION_ID` it is simple and fast 

```php
<?php

if (PHP_VERSION_ID < 80600) {
    enum SortDirection
    {
        case Ascending;
        case Descending;
    }
}
```

polyfill/SortDirection.php
```diff
@@ -0,0 +1,20 @@
...
+ * (c) Ignace Nyamagana Butera <nyamsprod@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+if (!enum_exists('SortDirection', false)) {
+    enum SortDirection
```
nyamsprod, line 15: Do not forget to update the `autoload.php` file in the root directory so that a standalone download of the package without composer can still work and autoload the Enum in PHP8.6-

composer.json
```diff
@@ -45,7 +45,10 @@
         "psr-4": {
             "League\\Csv\\": "src"
         },
-        "files": ["src/functions_include.php"]
+        "files": [
+            "src/polyfill86.php",
```
nyamsprod: I'd rather use a classmap and a different directory like polyfill at the same level as src to include the Polyfill there. If you add the SortDirection Enum inside the src directory you are misleading the user in thinking this class is part of the League\Csv package whereas it is not ... it is a polyfill

polyfill/sort_direction.php
okorneliuk: If polyfill is already defined by another library in the project, we will probably get an exception of trying to redefine the enum. I prefer to keep this as it is.

nyamsprod: true then we need to apply both requirements then. I did it like this on another of my package:

- [the polyfill](https://github.com/bakame-php/tokei/blob/main/lib/SortDirection.php) 
- [the autoloader](https://github.com/bakame-php/tokei/blob/main/autoload.php)
- [the composer.json update](https://github.com/bakame-php/tokei/blob/main/composer.json#L64-L71)

The same should be done for the csv codebase.

docs/9.0/extensions/doctrine.md
```diff
@@ -54,14 +54,15 @@ This extension package contains:
 use Doctrine\Common\Collections\Criteria;
 use League\Csv\Doctrine as CsvDoctrine;
 use League\Csv\Reader;
+use SortDirection;
```
nyamsprod: The doctrine extension is deprecated, I would not change the documentation for it as it will not be updated. This change in documentation should be reverted

polyfill/sort_direction.php
okorneliuk: Agreed :+1:

nyamsprod:
@okorneliuk I will merge as is your feature and make sure to add your name in the release. But I may update the implementation if needed thanks for the PR and the work 👍🏾

PR #583, opened 2026-07-31
Fix EscapeFormula round-trip for fields starting with the escape character

gaoflow:
`EscapeFormula::unescapeRecord()` is documented and tested as the exact inverse of `escapeRecord()`, but the pair is not an inverse for fields that start with the escape character itself.

`escapeField()` only prepends the escape character to fields starting with a formula trigger. A field the user typed as `'=2+5` is left untouched, which makes the escape non-injective: `escapeRecord(["'=2+5"]) === escapeRecord(["=2+5"]) === ["'=2+5"]`. On read-back, `unescapeField()` strips the leading `'` whenever the second character is a trigger, so a legitimate `'=text` or `'@handle` field loses its leading quote. The documented write/read round-trip corrupts data.

The fix makes both halves operate on the escape prefix:

- `escapeField()` also prepends the escape character when the field already starts with it, so escaped fields are distinguishable from literal ones;
- `unescapeField()` strips exactly one escape prefix when the remainder starts with a formula trigger or with the escape character, which is the inverse of the new escaping rule. It now also handles multi-character escape strings, which previously never unescaped at all because the old check only compared the first byte.

The existing contracts are unchanged: `unescapeRecord(["'=2+5"])` still returns `["=2+5"]` and the reader examples in the docs behave the same.

Round-trip tests now cover every first-two-character regime (formula trigger, escape+trigger, escape+escape, escape+plain, plain text) with the default, custom and multi-character escapes.

Repro (current master):

```php
$ef = new EscapeFormula();
$record = ['id', '=2+5', "'=text you typed", "'@notmyhandle", 'normal'];
$writer = Writer::fromString();
$writer->addFormatter($ef->escapeRecord(...));
$writer->insertOne($record);
$reader = Reader::fromString($writer->toString());
$reader->addFormatter($ef->unescapeRecord(...));
var_export($reader->first() === $record); // false: '=text you typed / '@notmyhandle lose their leading '
```

Full suite, phpstan (level max) and php-cs-fixer are green locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ All Notable changes to `Csv` will be documented in this file
 
 ### Fixed
 
-- None
+- `EscapeFormula` round-trip now preserves fields starting with the escape character
 
 ### Remove
 
--- a/src/EscapeFormula.php
+++ b/src/EscapeFormula.php
@@ -21,6 +21,7 @@ use function array_fill_keys;
 use function array_keys;
 use function array_map;
 use function is_string;
+use function str_starts_with;
 
 /**
  * A Formatter to tackle CSV Formula Injection.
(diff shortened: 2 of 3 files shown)


nyamsprod (approved):

nyamsprod:
thanks for the PR and the improvement in the feature 👍🏾

PR #591, opened 2026-09-02
Reset the fields a DateTimeField format does not parse instead of taking them from the current time

lenamonj:
`(new DateTimeField('Y-m-d'))->parse('2024-01-01')` returns a date whose time is the wall clock at parse time, so the same input parses to a different value every second.

`DateTimeImmutable::createFromFormat` fills every field the format does not mention from the current time.

Append `|` to the format so unparsed fields reset to zero.

New test fails on master (`Failed asserting that two strings are identical`, time portion is the current time) and passes with the change; `composer phpunit`, `composer phpstan` and `composer phpcs` clean.

### DIFF
--- a/src/Schema/DateTimeField.php
+++ b/src/Schema/DateTimeField.php
@@ -191,7 +191,8 @@ final class DateTimeField extends FieldEvaluator implements Field
         }
 
         try {
-            $value = $this->outputClass::createFromFormat($this->format, $value, $this->timezone);
+            // `|` resets the fields the format does not parse, which createFromFormat otherwise fills from the current time.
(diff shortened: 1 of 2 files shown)


nyamsprod:
@lenamonj I am not quit sure about this fix since the format is a user input it is up to the user to decide whether he/she wants the time part and how it should be defined. You are adding an opinionated restriction IMHO

lenamonj:
Fair point. The format is the user's, and `|` or `!` belongs in it, not appended behind their back. Happy to close this, or reduce it to one sentence in the DateTimeField docs noting that fields the format does not mention come from the current time unless the format resets them. Your call.

nyamsprod:
I will close the PR. Thanks for submitting it.

PR #595, opened 2026-09-15
Fix typo in the error message for invalid header columns

AmandoVledder:


nyamsprod:
thanks for the typo fix.

